PR #18, opened 2022-06-28
fix(web-console): remove file separator symbol from `queryRaw` result

argshook:
QuestDB might return data which can include non printing symbols, like
[File Separator](https://unicode-table.com/en/001C/)

Such symbol is not a valid token for `JSON.parse`, so code like
`JSON.parse(stringWithFileSeparatorSymbol)` breaks

This change simply removes that symbol. It is a non-printing symbol so
it should not be displayed in the web console grid anyway.

### DIFF
--- a/packages/web-console/src/utils/questdb.ts
+++ b/packages/web-console/src/utils/questdb.ts
@@ -112,6 +112,17 @@ export type Options = {
   explain?: boolean
 }
 
+const parseJSON = <T extends object>(raw: string): T | null => {
+  try {
+    // remove file separator symbol, as it's not valid a token for JSON.parse
+    return JSON.parse(raw.replace(/\u001C/g, ""))
+  } catch (e) {
+    console.error("Unable to parse JSON returned from QuestDB")
+    console.error(e)
+    return null
+  }
+}
+
 export class Client {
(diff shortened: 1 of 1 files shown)


puzpuzpuz:
Maybe it's better for us to remove illegal chars on the server side? This way not only Web Console, but any HTTP API client would be able to parse the result set. @bziobrowski WDYT?

I'd also propose checking other potentially problematic chars (maybe codepoints U+D800 to U+DFFF? something else?)

argshook:
> Maybe it's better for us to remove illegal chars on the server side? This way not only Web Console, but any HTTP API client would be able to parse the result set. @bziobrowski WDYT?
> 
> I'd also propose checking other potentially problematic chars (maybe codepoints U+D800 to U+DFFF? something else?)

me and @bziobrowski had a two sentence discussion about removing such symbols on server, but concluded that this QuestDB should not do this, as strings can contain arbitrary symbols. Removing things might be unexpected by the user.

On the other hand, since web console only previews data, any non-printing chars can be removed here.

bziobrowski:
Closing PR because I added nonprintable char encoding in JSON responses returned by backend .

PR #23, opened 2022-08-22
web-console: add note about CSV import throught COPY command

argshook:
This PR updates web console import dialog with a note about `COPY`
command

image

### DIFF
--- a/packages/web-console/src/partials/import.hbs
+++ b/packages/web-console/src/partials/import.hbs
@@ -28,61 +28,52 @@
         <div id="import-top" class="ibox-content">
           <div id="dragTarget" class="drag-target drag-idle">
             <div>
-              <i class="fa fa-copy fa-3x" style="margin: 0 15px 20px 0;"></i>
+              <i class="fa fa-copy fa-3x" style="margin: 0 15px 20px 0"></i>
               <i
                 class="fa fa-folder-open fa-3x"
-                style="margin: 0 15px 20px 0;"
+                style="margin: 0 15px 20px 0"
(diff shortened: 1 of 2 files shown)


packages/web-console/src/partials/import.hbs
```diff
@@ -28,61 +28,51 @@
...
-                style="margin: 0 15px 20px 0;"
+                style="margin: 0 15px 20px 0"
               ></i>
+              <i class="fa fa-file-alt fa-3x" style="margin: 0 15px 20px 0"></i>
             </div>
-            <h2 style="font-weight: 500;">
-              Drag files or paste data here to
-              import
+            <h2 style="font-weight: 500">
+              Import CSV file
```
amyshwang, line 39: Seeing the screenshot, I wonder if this line is needed at all. 
Should we consider changing this to "click here to browse" with the link?

amyshwang (commented):
Some more thoughts but overall looks good.

packages/web-console/src/partials/import.hbs
argshook: I actually tried exactly that, but it felt a bit empty, I feel like some explanation of a CTA makes sense.

Wdyt?

argshook: image

amyshwang: Actually you're right! I think the original one looks better!

amyshwang (approved):
Based on the screenshot it looks good.

PR #28, opened 2022-09-22
web-console/Add shortcuts widget

insmac:
People are not aware of the keyboard shortcuts available, as they are not well documented and/or require using certain features through context menu to see the keyboard alternative.

This widget lists the most important items available both globally and within SQL editor area.

CleanShot 2022-09-22 at 15 49 00@2x


packages/web-console/src/scenes/Editor/Shortcuts/index.tsx
```diff
@@ -0,0 +1,159 @@
+import React from "react"
+import styled from "styled-components"
+import { color } from "../../../utils"
+import { Text } from "../../../components"
+import { platform } from "../../../utils"
+
+type ShortcutsList = { keys: string[][]; title: string }[]
```
argshook, line 7: how come this is not clashing with https://github.com/questdb/ui/pull/28/files#diff-863cb95c15f7ab7369c679049c640042f4833c05f55ad42b09dd8ae13f474101R122

argshook (approved):

packages/web-console/src/scenes/Editor/Shortcuts/index.tsx
insmac: TypeScript magic from what I remember 🪄

argshook: perhaps lets avoid that magic?

cause reading this i need a sec to comprehend wtf https://github.com/questdb/ui/pull/28/files#diff-863cb95c15f7ab7369c679049c640042f4833c05f55ad42b09dd8ae13f474101R122-R128

argshook (approved):

PR #33, opened 2022-09-26
Do not duplicate existing queries when appending via URL param

insmac:
Detect if a query passed via URL param already exists in the editor buffer and instead of duplicating, highlight the first match.

Tests - to be added

CleanShot 2022-09-26 at 19 44 21@2x

### DIFF
--- a/packages/web-console/src/scenes/Editor/Monaco/index.tsx
+++ b/packages/web-console/src/scenes/Editor/Monaco/index.tsx
@@ -203,9 +203,18 @@ const MonacoEditor = () => {
 
     // Insert query, if one is found in the URL
     const params = new URLSearchParams(window.location.search)
-    const query = params.get("query")
+    // Support multi-line queries (URL encoded)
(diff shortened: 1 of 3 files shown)


packages/web-console/src/scenes/Editor/Monaco/index.tsx
```diff
@@ -203,9 +203,18 @@ const MonacoEditor = () => {
 
     // Insert query, if one is found in the URL
     const params = new URLSearchParams(window.location.search)
-    const query = params.get("query")
+    // Support multi-line queries (URL encoded)
+    const query = params.get("query")?.replaceAll("%0A", "\n")
```
argshook, line 207: @insmac why was this necessary? `URLSearchParams.prototype.get` should already return decoded string. Maybe it didn't work for you on some browser?

argshook: we double checked, it's not needed

PR #82, opened 2023-02-08
fix: delayed grid resize and column value overspill

### LINKED ISSUE #80: UI: truncated grid values spill into other columns
When column becomes to narrow to hold value, it spills into column on the left . 

grid_issue

### LINKED ISSUE #81: web-console prints `undefined` in grid
with `@questdb/web-console@0.0.11` we can observe a lone `undefined` printed in the grid area, after running a simple query.

In screenshot below an `ed` can be visible after the first column is shown.

Screenshot 2023-02-08 at 19 23 01

likely an issue somewhere in [grid.js](https://github.com/questdb/ui/blob/web-console/add-jest/packages/web-console/src/js/console/grid.js)

### LINKED ISSUE #83: Web console results have weird behaviour when running a query with the 'Chart' tab active
### Describe the bug

With QuestDB 7.0.1 whenever I run a query while having the 'Chart' tab active and if I go back to the 'Grid' Tab, results are represented only by a thin column. On expanding the column I can see data, and I need to them expand all the other columns. 

See the screenshots on next box

### To reproduce

Start QuestDB 7.0.1

Navigate to the web console, select the Grid tab, do a `select *` on any table

image

Now change to the Grid tab. Oh no!

image

You can just start dragging columns, but it shouldn't be needed

image

If you run the same query from the Grid tab, everything looks good

image

### Expected Behavior

On previews versions you could just run a query on the Chart tab and switch to Grid and it would work. Screenshot from the public demo site running QuestDB 7.0.0

image

### Environment

```markdown
- **QuestDB version**: 7.0.1
- **OS**: MacOS 13.1
- **Browser**: Chrome
```

### Additional context

_No response_

bluestreak01:
Fixes #83
Fixes #80
Fixes #81

### DIFF
--- a/packages/web-console/src/js/console/grid.js
+++ b/packages/web-console/src/js/console/grid.js
@@ -525,7 +525,6 @@ export function grid(root, msgBus) {
         // handle.innerHTML = i
         handle.onmousedown = columnResizeStart
         header.append(handle)
-
         const hBorderSpan = document.createElement('span')
         hBorderSpan.className = 'qg-header-border'
         h.append(hBorderSpan)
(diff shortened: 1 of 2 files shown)


argshook:
Thanks! i can confirm this fixes #81

argshook (approved):
👍 

also learned new word: hysteresis

PR #86, opened 2023-02-09
chore: improve grid render performance

bluestreak01:
## Motivation

To prepare grid for column movement and column freezing I moved grid render from stylesheets to element styles. This in itself improved grid scrolling performance.

### Before

image

### After

image

## Look and feel changes

- added subtle vertical lines to separate cell values
- removed cell hover to avoid ambiguously highlighting multiple grid cells
- made row hover background more subdued
- there is border on the right edge of the grid to make look and feel consistent

image


packages/web-console/src/js/console/grid.js
```diff
@@ -603,12 +572,12 @@ export function grid(root, msgBus) {
...
 
   function removeClass(e, className) {
-    e.className = e.className.replace(className, '')
+    e.className = e.className.replace(className, '').trim()
   }
 
   function addClass(e, className) {
     if (e && !e.className.includes(className)) {
-      e.className += className
+      e.className += ' ' + className
```
argshook, line 580: no need to change anything, just want to share info about `classlist`, might be helpful for lines like this.

https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

argshook (approved):
This is awesome!

packages/web-console/src/js/console/grid.js
bluestreak01: hmm, do that specifically did not work for me, let me retry

bluestreak01: ah, `e.classList.add()` - thanks!

bluestreak01: addressed in https://github.com/questdb/ui/pull/86/commits/9c85caacdeda3c70a6465cdb5caab1b9df8f4565

bluestreak01:
also highlighted nulls

image

packages/web-console/src/styles/_grid.scss
```diff
@@ -113,10 +131,11 @@ $visible-drag-handle-margin: 10px;
 .qg-drag-handle {
   // place drag handle so that visible "line" is in the middle of it
   // to add hysteresis for the drag cursor
-  margin-left: -$visible-drag-handle-margin / 2;
-  width: $visible-drag-handle-margin * 2 !important;
+  margin-left: -$visible-drag-handle-margin - $drag-handle-margin;
+  width: $visible-drag-handle-margin * 2 ;
   height: $row-height;
```
argshook, line 136:
```suggestion
  width: $visible-drag-handle-margin * 2;
  height: 100%;
  top: 0;
```

argshook (changes requested):

packages/web-console/src/styles/_grid.scss
argshook: without this change, the drag handle position is off:

Screenshot 2023-02-13 at 15 12 17

PR #87, opened 2023-02-10
chore(wal): add wal telemetry to telemetry tracking

ideoma:
Push WAL write telemetry as aggregated events with different origins

- `origin=1000` total row count written to WAL tables in thousands
- `origin=2000` total commits written to WAL tables
- `origin=3000` max commit write latency in seconds

WAL telemetry is aggregated by hour, so there will be max 3 rows per every hour.
All values are capped to max short 32767 value.

WAL telemetry is described at https://github.com/questdb/questdb/pull/2930

### DIFF
--- a/packages/web-console/src/store/Telemetry/epics.ts
+++ b/packages/web-console/src/store/Telemetry/epics.ts
@@ -132,13 +132,43 @@ export const startTelemetry: Epic<StoreAction, TelemetryAction, StoreShape> = (
       if (remoteConfig?.lastUpdated) {
         return from(
           quest.queryRaw(
-            `SELECT cast(created as long), event, origin
-                FROM ${TelemetryTable.MAIN}
-                WHERE created > '${new Date(
-                  remoteConfig.lastUpdated,
-                ).toISOString()}'
-                LIMIT -10000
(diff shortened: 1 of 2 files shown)


packages/web-console/src/store/Telemetry/epics.ts
```diff
@@ -132,13 +132,43 @@ export const startTelemetry: Epic<StoreAction, TelemetryAction, StoreShape> = (
...
-                  remoteConfig.lastUpdated,
-                ).toISOString()}'
-                LIMIT -10000
-            `,
+            `with tel as (
+             SELECT cast(created as long), event, origin
+             FROM ${TelemetryTable.MAIN}
+             WHERE created > '${new Date(
+              remoteConfig.lastUpdated,
+             ).toISOString()}'
```
argshook, line 140: `new Date` part is repeated throghout, would be nice to define it once and reuse

argshook (approved):

packages/web-console/src/store/Telemetry/epics.ts
ideoma: Ok, changed

PR #104, opened 2023-02-15
fix(grid): fix storage bug, which could manifest as exception in incognito browsers

bluestreak01:


argshook (approved):
Thanks!

Looks like this change fixes the test failure caught by tests on  https://github.com/questdb/ui/pull/103

PR #107, opened 2023-03-02
feat(grid): freeze columns on the left

bluestreak01:
- provides visual tools to freeze left columns
- rendering bugfixes


bluestreak01:
I would be keen to make release out of this

Issue #116, opened 2023-03-31
No information if a file being uploaded has been processed already


insmac:
### Describe the bug

If user tries to upload a file that has been already processed, the process fails (there is a REST `/chk` request to check the file before uploading), which is as intended, but no feedback is given in the UI.a 

### To reproduce

1. Go to the `Import` page, a second icon in the navbar
2. Upload a CSV file.
3. Try uploading the same file again.

### Expected Behavior

UI feedback stating that the file already exists.

### Environment

```markdown
- **QuestDB version**: 7.0.1
- **OS**: macOS Ventura 13.2.1
- **Browser**: all
```

insmac:
Issue moved to the dedicated UI repository (https://github.com/questdb/ui/issues)

PR #119, opened 2023-03-31
fix(web-console): Import UI fixes

insmac:
Bottom line: The entire Import page needs to be moved into React and adjusted visually to match the look and feel of the application. Until then - quick fixes are to make the page fully work.

Fixes:
[Table schema not refreshed after successful CSV upload](https://github.com/questdb/ui/issues/115)
[No information if a file being uploaded has been processed already](https://github.com/questdb/ui/issues/116)
[CSV upload is broken in Web Console](https://github.com/questdb/ui/issues/118) (drag and drop issue)

Drag and drop screen recording - notice how the schema is refreshed and grid contains resuls from a recently uploaded file.

https://user-images.githubusercontent.com/1871646/229133419-ac4352eb-b52f-4e79-989d-de09ebdc2be0.mp4

Notable changes:
- Removed the splitter to resize panes. It creates problem by not refreshing properly when a new file status row is added, thus not displaying content ([i.e. if a file already exists, has been uploaded etc](https://github.com/questdb/ui/issues/11)).
- Fixed some issues that happened when the drag issue has been cleared out (pushing DOM elements to an array).


insmac:
CleanShot 2023-04-03 at 12 47 46@2x
CleanShot 2023-04-03 at 12 47 06@2x

@bluestreak01 I have fixed found issues (as per Slack thread):
- status labeling
- file upload info (error, success, etc - as per attached screenshot)
- actions buttons

Much more can be done, but because I'll need to rewrite the logic and UI for React anyway, it would be twice the work (since the current solution is incompatible), I would want to pause at this point and ideally merge since this PR fixes a number of bugs in the current release, so there is a big net benefit in my view.

insmac:
Closing in favor of https://github.com/questdb/ui/pull/130

PR #133, opened 2023-04-28
feat(web-console): Add the ability to copy formatted error/success messages to clipboard

MatthewCallis:
Add the ability to copy formatted error/success messages to clipboard, resolves https://github.com/questdb/ui/issues/42.

Open Questions:
- [x] `Add an ability to copy the formatted message to clipboard`, what does formatted mean in this context? How much of the notification do we want to copy? I saw formatted and assumed the query itself so far.
- [x] There are dueling copy to clipboard methods, one deprecated one in `packages/web-console/src/utils/copyToClipboard.ts` and one in `packages/web-console/src/js/console/grid.js` that lacks `.catch()`.
- [x] `Add an additional copy icon - the one we have next to QuestDB version number`, it isn't clear where this is and the closest match appears to be `packages/web-console/src/scenes/Footer/BuildVersion/index.tsx` but there are only 2 icons, and neither are appropriate for a copy action as both already have clear functions: `UpgradeIcon` (styled `ArrowUpCircle`), `ExternalLink`; for the sake of this project I am using the `ArrowUpCircle` icon for now; a clipboard or the double square icon (used by GitHub) is typically used to represent a copy action.
- [x] I cannot find generalized tests in the repo, what would be the most appropriate place to add test cases for this?
- Misc. Bug: Context menu on table schema isn't clear as the built in browser menu hides it.

Example Tooltip & Icon:

https://user-images.githubusercontent.com/67945/235800189-6d8bd11e-33cf-49b0-9dec-962db67317ee.mov


argshook:
Hey @MatthewCallis, thanks for contributing.
I address your questions below. Feel free to follow-up if you'd need!

>     * [ ]  `Add an ability to copy the formatted message to clipboard`, what does formatted mean in this context? How much of the notification do we want to copy? I saw formatted and assumed the query itself so far.

Formatted essentially is readable. We would like it to be easy for users to quickly copy-paste notifications from logs to, say, slack. SQL in the log is one part of notification and arguably the error message is more important. But you're free to implement whatever UX you think would fit the best.

Consider a scenario where user is facing an issue and wants to share log entry to some support channel (could even be our public QuestDB slack).

>     * [ ]  There are dueling copy to clipboard methods, one deprecated one in `packages/web-console/src/utils/copyToClipboard.ts` and one in `packages/web-console/src/js/console/grid.js` that lacks `.catch()`.

I would suggest to use [`src/utils/copyToClipboard.ts`](https://github.com/questdb/ui/blob/main/packages/web-console/src/utils/copyToClipboard.ts) because we plan to externalize `grid.js` in future and so it would be good not to introduce a new dependency on grid at the moment.

>     * [ ]  `Add an additional copy icon - the one we have next to QuestDB version number`, it isn't clear where this is and the closest match appears to be `packages/web-console/src/scenes/Footer/BuildVersion/index.tsx` but there are only 2 icons, and neither are appropriate for a copy action as both already have clear functions: `UpgradeIcon` (styled `ArrowUpCircle`), `ExternalLink`; for the sake of this project I am using the `ArrowUpCircle` icon for now; a clipboard or the double square icon (used by GitHub) is typically used to represent a copy action.

If current used icons don't fit, you can pick one on https://styled-icons.dev/
we already have `styled-icons` dependency installed, so it should be good to go.

>     * [ ]  I cannot find generalized tests in the repo, what would be the most appropriate place to add test cases for this?


We have a few cypress browser tests in [`packages/browser-tests`](https://github.com/questdb/ui/blob/main/packages/browser-tests/README.md) folder. Check the readme there or [readme of web-console package](https://github.com/questdb/ui/blob/main/packages/web-console/README.md#L91) to see how to run them.

>     * Misc. Bug: Context menu on table schema isn't clear as the built in browser menu hides it.

Perhaps you could open a new issue for the bug you mentioned? From this description i'm not sure i can reproduce.

packages/web-console/src/scenes/Notifications/Notification/SuccessNotification/index.tsx
```diff
@@ -41,22 +48,61 @@ const ZapIcon = styled(Zap)`
...
+
 export const SuccessNotification = (props: NotificationShape) => {
-  const { createdAt, content, sideContent, jitCompiled } = props
+  const { createdAt, content, sideContent, jitCompiled, request, result } = props
+  const handleCopyToClipboard = useCallback(() => {
+    let clipboardText = `[${format("HH:mm:ss", createdAt)}] ${request?.query}\n`
+    if (result?.timings) {
+      try {
+        const { count, compiler, fetch, execute } = result.timings;
+        clipboardText += `✔ Success: ${result.count?.toLocaleString()} rows in ${formatTiming(fetch)}\n`;
```
argshook, line 62: to be honest, i'm not really sure if copy to clipboard should format as a table. Perhaps it's enough to copy same text that is visible in UI (except not truncate it). WDYT?

nonetheless, if we'd continue with formatting as a table, then i'd suggest to not use a dependency but build a util.

this should do the trick:

```
const row1 = ["Execute", "Network", "Total", "Count", "Compile"];
const row2 = ["6.07ms", "86.93ms", "93ms", "0", "515.97µs"];
const rows = [row1, row2];

const renderTable = (rows) => {
  const colMax = rows[0].map((_, i) =>
    Math.max(...rows.map((row) => String(row[i]).length))
  );

  const header = `| ${rows[0]
    .map((col, i) => col.padEnd(colMax[i]))
    .join(" | ")} |`;

  const separator = `+${colMax.map((max) => "-".repeat(max + 2)).join("+")}+`;

  const data = rows
    .slice(1)
    .map(
      (row) =>
        `| ${row.map((col, i) => String(col).padEnd(colMax[i])).join(" | ")} |`
    );

  return [separator, header, separator, ...data, separator].join("\n");
};

console.log(renderTable(rows));

/*

+---------+---------+-------+-------+----------+
| Execute | Network | Total | Count | Compile  |
+---------+---------+-------+-------+----------+
| 6.07ms  | 86.93ms | 93ms  | 0     | 515.97µs |
+---------+---------+-------+-------+----------+

*/

```

cc @insmac wdyt?

packages/web-console/src/scenes/Notifications/Notification/SuccessNotification/index.tsx
```diff
@@ -41,22 +48,61 @@ const ZapIcon = styled(Zap)`
...
-          placement="top"
-          tooltip="JIT Compiled"
-        />
-      ) : (
-        <CheckmarkOutlineIcon size="18px" />
-      )}
-      <Content>{content}</Content>
-      <SideContent>{sideContent}</SideContent>
-    </Wrapper>
+    <PopperHover
```
argshook, line 92: i think if tooltip is added, then it should only be shown when icon is hovered, not the entire row.

please note this little button at the bottom right, which allows to expand the notifications area. When expanded, the current tooltip UX is not good IMO.

Screenshot 2023-05-03 at 14 49 00

argshook (changes requested):

packages/web-console/src/scenes/Notifications/Notification/SuccessNotification/index.tsx
MatthewCallis: I've updated it such that the tooltip now shows to the right and is before the timestamp:

Screenshot 2023-04-28 at 3 43 38 PM

packages/web-console/src/scenes/Notifications/Notification/SuccessNotification/index.tsx
MatthewCallis: I'm not sure what the most common use case would be for copying the data out. I have updated it to use this code rather than the package. If it is for debugging it may be nice to pre wrap it in `code` backticks for Slack or the like.

packages/browser-tests/cypress.config.js
```diff
@@ -1,4 +1,5 @@
 const { defineConfig } = require("cypress");
+const clipboardy = require("clipboardy");
```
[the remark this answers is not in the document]
MatthewCallis: Switched to an inline method in the test file itself.

packages/web-console/src/scenes/Notifications/Notification/SuccessNotification/index.tsx
insmac: @argshook 

I think `Copy to clipboard` should format in a way that's easy to read when pasted in channels like Slack or GitHub (when creating a new issue, for example).

For one, it's very hard to view partial screenshots (i.e. showing the notification log) on mobile, and even sometimes on desktop - even the thumbnails are small. 
Additionally, you can copy and paste further in a more convenient manner.
This follows the pattern of QuestDB REST API showing operation status (like after importing a CSV file) as an ASCII table by default, unless JSON format is requested 🙂 

In short: I like this approach!

PR #143, opened 2023-05-16
web-console: Added Statement Timeout Configuration and retry button for timeout

aikchun:
Added an extra param `isFromError` for `toggleRunning` to inform the query run in `Monaco` to run the query that is set from the `ErrorNotification

Added an extra param to `queryRaw` to set headers for the query.

Added a RetryButton for ErrorNotifcation that are timeouts.

Clicking on the RetryButton will run the query, and also based on the `running.value` set the button to disabled.
As a general heuristic the query running this way will be doubling the timeout value.

Initial Request with timeout at 6000
timeout-6000

Button disabled while a query is running
Screenshot 2023-05-16 at 9 03 07 PM

Query made with retry button with timeout at 12000
Screenshot 2023-05-16 at 9 03 29 PM

Edit:
UX improvements:
Made the retry button green with greyed out background to increase prominence
Screenshot 2023-05-19 at 7 41 51 PM

When running, the button turns to a stop button that can be used to stop the query if the user wishes.
Screenshot 2023-05-19 at 7 43 34 PM

Some further UX improvements that can be made/considered:
- Automatic retrying of query with increased timeout.
- If retrying query still causes a timeout, instead of adding another timeout notification, we can check if the last notifications, is a timeout notification with same query, if so, we can either replace or skip adding the notification.

Some of the best practices for react that I am familiar with are:
- Use functional components as much as possible (applied in PR)
- If a state change causes a dispatch, use the dispatch in a `useEffect`. (applied in PR)
- Use multiple `useEffect` to break up the side effects based on their dependencies rather than putting them all in one (applied in PR)
- Even though `useState(heavyComputation())` is initialized once, `heavyComputation()` will still be invoked everytime the component renders, use `useMemo` to improve performance.

### DIFF
--- a/packages/web-console/CHANGELOG.md
+++ b/packages/web-console/CHANGELOG.md
@@ -14,6 +14,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `Fixed` for any bug fixes.
 * `Security` in case of vulnerabilities.
 
+## 0.2.1 - 2023-05-18
+
+### Added
+
+- Allow editing schema on existing tables (column type, designated timestamp and its pattern, geohash precision) [#140](https://github.com/questdb/ui/pull/140)
+- Adjust pasting CSV files from clipboard to not require page wrapper focus [#147](https://github.com/questdb/ui/pull/147)
+
+### Changed
+
+- Simplify and refactor state management in schema editor [#139](https://github.com/questdb/ui/pull/139)
+
 ## 0.0.11 - 2023-02-13
 
 ### Fixed
--- a/packages/web-console/package.json
+++ b/packages/web-console/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@questdb/web-console",
-  "version": "0.2.0",
+  "version": "0.2.1",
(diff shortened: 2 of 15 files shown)


bluestreak01:
Hi Aik,

There are couple of issue with the PR:
- I would like to see you apply apply best practices you're familiar with to the PR description
- It would be good if you attempt to improve the UX of the feature you're adding in this PR

Thanks,
Vlad

aikchun:
> Hi Aik,
> 
> There are couple of issue with the PR:
> 
>     * I would like to see you apply apply best practices you're familiar with to the PR description
> 
>     * It would be good if you attempt to improve the UX of the feature you're adding in this PR
> 
> 
> Thanks, Vlad

@bluestreak01 Hi there. Thanks for the feedback.

I have made some changes to the PR and updated the PR description.
Do let me know if further improvements can be made for this PR or in another.

Thank you
Aik Chun

goodroot:
Hi there - sorry about that, but this went long.

We've decided to close it as state is now far away from when these changes were propsoed.

PR #154, opened 2023-05-30
chore(web-console): Import UI fixes + timestamp input update

insmac:
- Fixed an issue when the user could not paste text (i.e. timestamp format) into form inputs because this action would close them and reset their state.
- Fixed an issue when pasting a file to an already existing list would overwrite the queue.
- Added a sample timestamp format text that stays below the pattern input all the time.
CleanShot 2023-05-30 at 16 58 45@2x

### DIFF
--- a/packages/web-console/src/scenes/Import/ImportCSVFiles/table-schema/column.tsx
+++ b/packages/web-console/src/scenes/Import/ImportCSVFiles/table-schema/column.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import { Form } from "../../../../components/Form"
-import { IconWithTooltip } from "../../../../components"
+import { IconWithTooltip, Text } from "../../../../components"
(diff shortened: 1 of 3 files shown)


packages/web-console/src/scenes/Import/ImportCSVFiles/table-schema/column.tsx
```diff
@@ -93,50 +94,53 @@ export const Column = ({
...
-              >
-                Designated
-              </Button>
-            }
-            tooltip="Mark this column as a designated timestamp"
-            placement="top"
-          />
-        </Controls>
+          </Controls>
+          <Text color="comment">Example: {DEFAULT_TIMESTAMP_FORMAT}</Text>
```
argshook, line 142: personal taste, i'd use `gray2`

argshook (approved):

PR #162, opened 2023-06-05
feat(web-console): Feedback dialog

insmac:
This PR introduces a new feature, Feedback dialog, utilizing the same UI and logic as in QuestDB Cloud, and using Cloud's feedback API to route the message to the team.

CleanShot 2023-07-25 at 17 21 57@2x

Base QuestDB configuration specs are sent as payload in addition to the message itself.


packages/web-console/src/scenes/Editor/Menu/index.tsx
```diff
@@ -278,6 +286,43 @@ const Menu = () => {
 
       <Separator />
 
+      <FeedbackDialog
+        withEmailInput
+        isSubmitting={isFeedbackSubmitting}
```
argshook, line 291: feels like `isSubmitting` should be handled internally within `FeedbackDialog`

packages/web-console/src/scenes/Editor/Menu/index.tsx
```diff
@@ -278,6 +286,43 @@ const Menu = () => {
...
+        title="Web Console feedback"
+        subtitle="Let us know your thoughts"
+        trigger={({ setOpen }: { setOpen: (isOpen: boolean) => void }) => (
```
argshook, line 294:
```suggestion
        trigger={({ setOpen }) => (
```

should be inferred properly by typescript

argshook (changes requested):
Nice work!

I want to point a few things worth considering:

# 1. resizable `textarea`

when resized, user can get this layout:



would be good to prevent this by tinkering with:
* `min-width`/`max-width`
* `min-height`/`max-height`

# 2. Inputs with labels

i would suggest a slightly different layout:

| current | suggested |
| --- | --- |
|  |  |

changes:

1. add `Email` and `Message` labels before input fields
2. use `placeholder` for input value example, not as a label of the field. More ramblings about it if you're interested: [Placeholder Research - Low Vision Accessibility Task Force](https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/Placeholder_Research#Placeholder_Attribute_Is_Not_A_Label.21)
3. shorter `Optional, ...` message and in brighter color. I find the blueish color too dark to read
4. wider gap between two input fields

# 3. keep values of input fields even when dialog is closed

consider a scenario:

1. user wants to give feedback
1. click feedback, type message
1. forget some detail, want to see it, but feedback dialog is on top of the UI, hiding something important
1. user clicks x, dialog is closed
1. they check that detail and want to continue writing the feedback message
1. click feedback again, and previous message is gone, rage quit

argshook (approved):
Great work, looks awesome!

PR #172, opened 2023-07-24
feat(react-components): Add `FeedbackDialog` and few others

insmac:
Changelog:
- Added `<FeedbackDialog>` component that can be reused between Web Console and the Cloud in order to provide a unified user experience. Replaced `react-hook-form` with a vanilla form handling, keeping JOI validation, swapped icons for `bootstrap`, in line with recent changes in `saas-client`.
- Added `<Box>` component.
- Added custom SVG icons.
- Added `<TextArea>` to be used outside of `react-hook-form` (similar to `<Input>`).


packages/react-components/src/components/FeedbackDialog/types.ts
```diff
@@ -0,0 +1,13 @@
+export type Category =
```
argshook, line 1: is this file used anywhere?

packages/react-components/src/components/icons/grafana.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
```
argshook, line 1: is this icon used anywhere?

packages/react-components/src/components/icons/postgresql.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
```
argshook, line 1: is this icon used anywhere?

packages/react-components/src/components/icons/grafana.tsx
insmac: It's used here: https://github.com/questdb/saas-client/blob/master/modules/connection-details/badges.tsx#L6 (so we can refect to this package in Cloud UI as well)

packages/react-components/src/components/icons/postgresql.tsx
insmac: Same as above: https://github.com/questdb/saas-client/blob/master/modules/connection-details/badges.tsx#L7

packages/react-components/src/components/FeedbackDialog/types.ts
insmac: Good catch, not needed, as it's part of the API!

packages/react-components/src/components/icons/grafana.tsx
argshook: ok! i thought the scope of this PR is just feedback dialog, didn't realise you're also extracting icons.

cool tho

insmac: @argshook That's because one of the icons is used in Cloud's Feedback Dialog, therefore I used it here as well :)

argshook (approved):

PR #173, opened 2023-07-26
feat(react-components): Add optional `email` field to `FeedbackDialog`

insmac:
Add an optional email field to the feedback form, so that the team can reach back to Web Console users (Cloud Feedback has email in the payload by default).

### DIFF
--- a/packages/react-components/src/components/FeedbackDialog/index.tsx
+++ b/packages/react-components/src/components/FeedbackDialog/index.tsx
@@ -6,6 +6,7 @@ import { Card } from "../Card";
 import { ForwardRef } from "../ForwardRef";
 import { Loader } from "../Loader";
 import { Overlay } from "../Overlay";
+import { Input } from "../Input";
 import { TextArea } from "../TextArea";
 import { Text } from "../Text";
 import Joi from "joi";
@@ -14,12 +15,19 @@ import { Undo } from "../icons/undo";
 import styled from "styled-components";
 
 type Values = {
+  email: string;
   message: string;
 };
 
 const minLength = 20;
 const maxLength = 1000;
 const schema = Joi.object({
+  email: Joi.string()
+    .email({ tlds: { allow: false } })
+    .messages({
+      "string.base": "Please enter an email address, it is required",
+      "string.email": "Please enter a valid email address",
+    }),
   message: Joi.string()
     .min(minLength)
     .max(maxLength)
@@ -33,15 +41,26 @@ const schema = Joi.object({
     }),
 });
 
+const FormControl = styled.div`
+  display: grid;
+  gap: 1rem;
+  width: 100%;
+`;
+
 const ChatIcon = styled(Chat)`
   color: ${({ theme }) => theme.color.foreground};
 `;
 
 const StyledDialogContent = styled(AlertDialog.Content)`
+  display: grid;
+  gap: 1rem;
   background: #282a36;
 `;
 
 const StyledCardContent = styled(Card.Content)<{ withAfterMessage: boolean }>`
+  display: grid;
+  gap: 1rem;
+  width: 100%;
   ${({ withAfterMessage }) => !withAfterMessage && `padding-bottom: 0`}
 `;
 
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/FeedbackDialog/index.tsx
```diff
@@ -155,15 +183,17 @@ export const FeedbackDialog = ({
...
                 { message: e.target.message.value },
-                "message"
+                ["email", "message"]
               );
               if (Object.keys(errors).length === 0) {
                 onSubmit({
+                  email: withEmailInput ? e.target.email.value : undefined,
                   message: e.target.message.value,
                 });
+                setOpen(false);
```
argshook, line 196: i think `Props['onSubmit']` should require a Promise. This way `setOpen(false)` could be closed only when `onSubmit` resolves successfully.

In case of error, the dialog would still be open and user could see error message

packages/react-components/src/components/FeedbackDialog/index.tsx
```diff
@@ -194,10 +224,35 @@ export const FeedbackDialog = ({
...
+                    <Input
+                      name="email"
+                      type="email"
+                      placeholder="Your e-mail"
+                      autoFocus
+                    />
+                  )}
+                  {withEmailInput && (
+                    <Text color="comment">
+                      Providing an e-mail address is optional, but if you do so,
```
argshook, line 238: the more i think about this, the more i lean to idea that additional fields should be rendered via a render prop which is provided by user and not hardcoded in the component.

Consider a case where we want to change this text. Changing it would
involve PR and release of `@questdb/react-components` as well as a
release of the consumer code.

Maybe something like this instead?

```
<FeedbackDialog
  onSubmit={(values) => {
    return Promise.resolve()
    // or, in case of failure
    return Promise.reject({ errors: [] })
  }}>
  {(Field) => (
    <Field label="Your email" info="Providing an e-mail address is optional...">
      <Input type="email" name="email" />
    </Field>
  )}
</FeedbackDialog>
```

wdyt?

argshook (commented):
the current implementation in this PR will work and should hold for a bit, but being conscious of the movements in UI, we might want to consider making this bit a bit more dynamic and component-y.

Throwing some ideas in comments

packages/react-components/src/components/FeedbackDialog/index.tsx
insmac: @argshook I thought about this and I like your idea, but there are two things I have issues with:

1. Fields are closely tied with API implementation in `saas-client`. My thinking is because of this, making the UI part dynamic will still involve changes & deployment of the Cloud UI, and therefore I'm not sure if it's worth future-proofing the code to that extent _with the current API implementation_. 
2. Render prop would mean that we need to make sure rendering is aligned with the rest of the `<FeedbackDialog`>` - in the current, simple solution, we know that it will be displayed the way is meant to be (i.e. no accidental positioning / width / padding issues).

packages/react-components/src/components/FeedbackDialog/index.tsx
```diff
@@ -14,12 +15,19 @@ import { Undo } from "../icons/undo";
...
 type Values = {
+  email: string;
   message: string;
 };
 
 const minLength = 20;
 const maxLength = 1000;
 const schema = Joi.object({
+  email: Joi.string()
+    .email({ tlds: { allow: false } })
```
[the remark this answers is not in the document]
insmac: My understanding is that if I don't pass `required()` explicitly, the schema is still valid when `withEmailInput` is false, at least I've been able to get to the request phase, but I'll double check.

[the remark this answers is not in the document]
insmac: Confirmed that `allow(undefined)` is not required if the field is not explicitly defined as `.required()`

packages/react-components/src/components/FeedbackDialog/index.tsx
argshook: okay, let's roll with it for now and we'll see when or if we need to augment current code!

argshook (approved):

PR #175, opened 2023-07-28
chore(web-console): Remove Handlebars

insmac:
Remove the unnecessary template loader and resort to standard HTML.


argshook (approved):
lovely! :heart:

PR #179, opened 2023-07-31
chore(web-console): Assets cleanup

insmac:
This PR removes unused, outdated, or unnecessarily heavy assets and code:

- Remove Bootstrap, as it's no longer used except for small parts in Form, where it can be replaced with simple custom styling
- Remove unused imported SCSS files (buttons, old import styling, etc)
- Remove fontawesome package (this alone appears to shave off half of the CSS bundle)
- Remove unused CSS classes.

Cypress snapshot run yielded no diff at my local machine between `main` and this branch.

### DIFF
--- a/packages/web-console/CHANGELOG.md
+++ b/packages/web-console/CHANGELOG.md
@@ -16,17 +16,32 @@ and this project adheres to
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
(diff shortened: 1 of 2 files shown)


argshook:
nice, looks good! always great to reduce the weight of the bundle, great work!

i added two more screenshot tests for imports and settings page here: https://github.com/questdb/ui/pull/181

argshook (approved):

Issue #188, opened 2023-08-16
Designated timestamp check on create table UI weird behaviour


javier:
On the nes (7.3.1) UI for create table I see when I click on the designated word (but not on the checkbox), the label goes green, but the check is not marked. However, when I click on create the table actually gets the designated timestamp option for that column.

![image](https://github.com/questdb/ui/assets/3839/e4b4fb0f-9b8a-45d9-9c4d-9ee3ea13f073)

![image](https://github.com/questdb/ui/assets/3839/1050297f-cf8d-485f-be99-9f1f609d3d4a)

I would expect the checkbox to be activated consistently when the label is activated. Right now it is a weird inconsistent behaviour

Issue #189, opened 2023-08-16
Designated timestamp silently accepts multiple checks


javier:
The Create table UI accepts (with no validation errors) multiple timestamp columns to be marked as designated timestamps, producing a SQL statement where the last column is used

![image](https://github.com/questdb/ui/assets/3839/3d0c846c-2adb-49f2-9067-97b39b725c85)

![image](https://github.com/questdb/ui/assets/3839/2b6f82c9-68e4-4bd3-bab2-05703999025b)

We should probably prevent this behaviour

javier:
New types work fine. However maybe we'd prefer alphabetical order for the type drop down?

![image](https://github.com/questdb/ui/assets/3839/eab54c36-ae66-4cd5-8258-ccf96d2d32c1)

PR #191, opened 2023-08-22
fix(web-console): Fix timestamp-related Create Table UI issues

insmac:
Closes https://github.com/questdb/ui/issues/188, https://github.com/questdb/ui/issues/189

- Fix incorrect timestamp check icon when selected 
- Fix an error where multiple designated timestamps can be selected
- Add the missing data types: `ipv4`, `byte`, `binary`, `long256`, `uuid`.
- Fix an unnecessary second scrollbar appearing in the UI


javier:
I can still do this (with some dodgyness) 

image

Resulting on this 

```
![image](https://github.com/questdb/ui/assets/3839/ec68a424-f19a-459b-802a-82e4788acfce)
```

The sequence here is adding three columns with the same name and all of them with timestamp type. Then select any of those as designated. The three get marked as designated. Now rename the other two columns. Click on create. No errors

javier:
The new types work fine. However, maybe alphabetical order would be preferred for the type dropdown?

image

javier (approved):
It looks good to me!

PR #192, opened 2023-08-30
chore(web-console): support writing multiple files to the same target

insmac:
Allow writing to the same target table name from different files.

This PR makes sure each file added to the queue is treated as a separate entity and there are no glitches in the UI.

https://github.com/questdb/ui/assets/1871646/9eef49e2-d14d-4974-a0b3-fd0d5cb9abc8

### DIFF
--- a/packages/web-console/src/scenes/Import/ImportCSVFiles/utils.ts
+++ b/packages/web-console/src/scenes/Import/ImportCSVFiles/utils.ts
@@ -1,0 +1,49 @@
+import { SchemaColumn } from "components/TableSchemaDialog/types"
+
+export const isGeoHash = (type: string) => type.startsWith("GEOHASH")
+
+export const extractPrecionFromGeohash = (geohash: string) => {
+  const regex = /\(([^)]+)\)/g
+  const matches = regex.exec(geohash)
+  if (matches && matches.length > 1) {
+    return matches[1]
+  }
+  return ""
+}
+
+export const mapColumnTypeToUI = (type: string) => {
+  if (isGeoHash(type)) {
+    return "GEOHASH"
+  } else return type.toUpperCase()
+}
+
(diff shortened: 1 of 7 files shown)


packages/web-console/src/scenes/Import/ImportCSVFiles/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+  return column
+}
+
+const str = () =>
+  (
+    "00000000000000000" + (Math.random() * 0xffffffffffffffff).toString(16)
+  ).slice(-16)
+
+export const uuid = () => {
```
jerrinot, line 35: is this how you generate a random uuid in JS? 😱

insmac: @jerrinot Nope, there are libraries to do so, but they all require `crypto` to be available in the system - so either HTTPS or localhost. We had users that use neither (just plain `http://`) and the system failed for them. For that reason I have just used a simple good-enough method to generate a random identifier (it's not persisted anywhere). 
I could use `new Date().getTime()` as an alternative, but you can drop multiple files to be uploaded at the same time, so there is a possibility for a trouble, I think 🙂

jerrinot: thanks! 
I agree it's unimportant here, I was just curious.

jerrinot (approved):
I tested the change locally, it does what the PR says. 
I get the idea behind the change - generate unique IDs. but I am far from a front-ent expert to say it's the best possible solution. but it looks good to me. 

I stumbled upon another weirdness while importing a CSV, but that's unrelated to this PR and it's very likely caused by a server-side bug. I'm investigating that separately.

PR #194, opened 2023-09-07
feat(web-console): Add IndexedDB wrapper for storage

insmac:


packages/web-console/package.json
```diff
@@ -60,7 +62,8 @@
     "sql-formatter": "^4.0.2",
     "styled-components": "5.1.1",
     "styled-icons": "10.47.0",
-    "throttle-debounce": "2.2.1"
+    "throttle-debounce": "2.2.1",
+    "usehooks-ts": "^2.9.1"
```
argshook, line 66: is this dep needed? doesn't look like it's used directly. maybe needed as dependency for something else?

insmac: @argshook It's not used in here, but I think you have originally added it because of the tabs implementation. I can remove it for this particular PR, sure.

argshook: yeah, i think let's only what's needed. That PR of mine is almost 1 year old

argshook (approved):

insmac:
Tested for the automated migration from the localstorage content - it worked fine on a few different browsers.

PR #195, opened 2023-09-08
chore(web-console) Add react-component builds to instructions

jerrinot:
I have no idea if that's the best way. but just following the guide did not work for me, the web-console build was missing the react-components and I figured out this could help. It did.


argshook (approved):
thanks!
the instructions should be updated indeed. The `@questdb/react-components` has recently become an inter-dependency of `@questdb/web-console`, hence the need to build it for local use.

PR #198, opened 2023-09-12
fix(grid): fix render crash when grid is empty

bluestreak01:
fixes:

https://github.com/questdb/ui/assets/7276403/7b4b95cb-d69a-4e01-a3af-1dd34073b762


insmac (approved):
Fixed the issue.

insmac (approved):
Both conditions work

PR #199, opened 2023-09-14
feat(web-console): general UI refresh, new News UI, new CSV Import UI

insmac:
This PR introduces an overall UI/UX refresh for the Web Console, in addition to adding `News` feature.

| Before | After |
|--------|--------|
| console-before | console-after |

The list of changes, clockwise:

General
- [x] Remove the `Settings` page as it has a single switch that wasn't used. 

Top Navbar

- [x] Move Shortcuts, Feedback, Docs, and Slack buttons under a new `Help` dropdown.
- [x] Update the Search feature, and change the library used to display the modal and restyle to match the widget on the website.
- [x] Adjust the Search trigger, which is auto-generated by Algolia, to look less like an input and more like a button.

(New) Right sidebar

- [x] Add a right sidebar with the following features: Help, News and Create Table. News and Create Table are opened as side modals.

Import
- [x] Move the Import panel into the Result (grid/chart) panel and add a toggle button to switch between it and results (if applicable).
- [x] Change the UI of the Import feature, so that the upload dropbox acts as a zero state and fills in the panel, and then the queue is second.
- [x] Do not query the imported dataset immediately, rather add the "View data" button to do it on demand.

Schema tab
- [x] Remove the page switch below the logo and use the schema icon to toggle collapsing the tab. In the next iteration, this space will house switching between the table list and saved queries/query history panel.

Editor
- [x] Move `Notifications` to the bottom of the Editor, from their bottommost position.
- [x] Update the Notifications UI so that the collapsed (default) view displaying the latest notification is in line with the tables list.
- [x] Minor UI improvements for notification items (i.e. for performance numbers) - UX for this task is still WIP

Results
- [x] Add a zero state for the initial Console load.
- [x] Remove the `Chart` / `Grid` tab bar, and create two side switches for that feature. These appear only if the grid has the result object present.
- [x] Move the grid action icons to the side from the now-defunct tab bar. These icons are context-aware, i.e. they should only be visible if the chosen view type is the grid and not the chart.


packages/web-console/src/scenes/News/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+    const rotateDegree = i % 2 === 0 ? 15 : -15
+    const keyframePercent = (i / numKeyframes) * 100
+    animation += `
+      ${keyframePercent}% {
+        transform: rotate(${rotateDegree}deg);
+      }
+    `
+  }
+  return animation
+}
```
argshook, line 29: ooof, let's not. i'll submit a rewrite

packages/web-console/src/scenes/News/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+  ${({ $unread }) =>
+    $unread &&
+    `
+  animation: 5s linear swing infinite;
+  @keyframes swing {
+    ${swing()}
+  }
+  `}
+`
```
argshook, line 42: will need to add `keyframes` to `styled-components` import

```suggestion
const swing = keyframes`
  0%,
  50%,
  100% {
    transform: rotate(0);
  }

  10%,
  30% {
    transform: rotate(-5deg);
  }

  20%,
  40% {
    transform: rotate(5deg);
  }
`

const BellIcon = styled(Notification2)<{ $unread: boolean }>`
  color: ${({ theme, $unread }) => theme.color[$unread ? "red" : "foreground"]};

  ${({ $unread }) =>
    $unread &&
    `
  animation: ${swing} 1s infinite ease-in-out;
  transform-origin: 50% 0;
  `}
`
```

argshook: ignore this, please see #202

packages/web-console/src/scenes/News/index.tsx
argshook: please see #202

bluestreak01 (approved):

PR #201, opened 2023-09-28
chore(web-console): add hot module replacement

insmac:
Add hot reloader for both React and Redux, avoiding the full page reloads upon saving files. This, in turn, decreases CPU usage while working.

The approach for Redux persistence layer takes the current legacy Redux setup into consideration - it's a fairly temporary measure.


argshook:
closing in favor of #205

Issue #97, opened 2023-02-14
table list disappears after erroneous SQL execution


bluestreak01:
When you execute a SQL, which overflows proxy HTTP buffer. Such as this one:

```sql
select 
'a'||x,
'b'||x,
'c'||x,
'd'||x,
'e'||x,
'f'||x,
'i'||x,
'j'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
[...]

PR #204, opened 2023-10-09
fix(web-console): keep table when 431 raised

jerempy:
431 raised from too long query no longer removes table

closes: https://github.com/questdb/ui/issues/97

solution implemented: check for 431 for too large query and don't set error if so, since the only thing the error was doing in this component is conditionally rendering error message there shouldn't be side-effects

### DIFF
--- a/packages/web-console/src/scenes/Schema/index.tsx
+++ b/packages/web-console/src/scenes/Schema/index.tsx
@@ -55,7 +55,7 @@ import {
   VirtualList,
 } from "../../components"
 import { actions, selectors } from "../../store"
-import { color, ErrorResult } from "../../utils"
+import { color, ErrorResult, isServerError } from "../../utils"
 import * as QuestDB from "../../utils/questdb"
 import Table from "./Table"
 import LoadingError from "./LoadingError"
@@ -192,7 +192,9 @@ const Schema = ({
         }
       },
       (error) => {
-        setLoadingError(error)
+        if (isServerError(error)) {
+          setLoadingError(error)
+        }
       },
       () => {
         setLoading(false)
--- a/packages/web-console/src/utils/index.ts
+++ b/packages/web-console/src/utils/index.ts
@@ -23,6 +23,7 @@
  ******************************************************************************/
 
 export * from "./copyToClipboard"
+export * from "./isServerError"
 export * from "./fetch"
 export * from "./fromFetch"
 export * from "./questdb"
--- a/packages/web-console/src/utils/isServerError.ts
+++ b/packages/web-console/src/utils/isServerError.ts
@@ -1,0 +1,5 @@
+// isServerError takes api response and checks if the error is of type server error or timeout
+// returns true if server error type
+export const isServerError = (response: Response): boolean => {
+    return response.status === 408 || response.status >= 500;
+};
--- a/packages/web-console/src/utils/questdb.ts
+++ b/packages/web-console/src/utils/questdb.ts
@@ -21,6 +21,7 @@
  *  limitations under the License.
  *
  ******************************************************************************/
+import { isServerError } from "utils";
 import { BusEvent } from "../consts"
 import { TelemetryConfigShape } from "./../store/Telemetry/types"
 
@@ -77,6 +78,7 @@ type RawResult = RawDqlResult | RawDdlResult | RawErrorResult
 
 export type ErrorResult = RawErrorResult & {
   type: Type.ERROR
+  status: number
 }
 
 export type QueryRawResult =
@@ -328,17 +330,18 @@ export class Client {
       }
     }
 
-    const errorPayload = {
-      error: `QuestDB is not reachable [${response.status}]`,
-      position: -1,
-      query,
-      type: Type.ERROR,
+    const errorPayload: Record<string, string | number> = {
+      status: response.status,
+      error: response.statusText,
     }
 
-    bus.trigger(BusEvent.MSG_CONNECTION_ERROR, {
-      ...errorPayload,
-      status: response.status,
-    })
+    if (isServerError(response)) {
+      errorPayload.error = `QuestDB is not reachable [${response.status}]`;
+      errorPayload.position = -1;
+      errorPayload.query = query;
+      errorPayload.type = Type.ERROR;
+      bus.trigger(BusEvent.MSG_CONNECTION_ERROR, errorPayload)
+    }
 
     // eslint-disable-next-line prefer-promise-reject-errors
     return await Promise.reject(errorPayload)


insmac:
Hey @jerempy, thanks for contributing! 🤝 

Your solution is partially correct, in that whereas it clears the "QuestDB is unreachable" error in the table list, that error is still treated as a connection error within the QuestDB client class ([questdb.ts](https://github.com/questdb/ui/blob/main/packages/web-console/src/utils/questdb.ts)), resulting in this state:
CleanShot 2023-10-09 at 11 25 51@2x

This is what we want to avoid by _not_ triggering this specific clode block:
https://github.com/questdb/ui/blob/main/packages/web-console/src/utils/questdb.ts#L331-L341
```TypeScript
    const errorPayload = {
      error: `QuestDB is not reachable [${response.status}]`,
      position: -1,
      query,
      type: Type.ERROR,
    }

    bus.trigger(BusEvent.MSG_CONNECTION_ERROR, {
      ...errorPayload,
      status: response.status,
    })
```

To trigger a genuine "broken connection" response, kill QuestDB when the Web Console is opened and try running any query. You should receive something like this:
CleanShot 2023-10-09 at 11 57 06@2x

Let's make an assumption we're going to support `408 (request timeout)`, `502 (bad gateway)` and `504 (gateway timeout)`. 

I would approach it somewhat like this:
1. Define a single block of code when the timeout check exists. I would put it below the `try/catch/finally` request block, like so:
```TypeScript
    // detect a gateway timeout
    if (
      [408, 502, 504].includes(response.status)
    ) {
      const errorPayload = {
        error: `QuestDB is not reachable [${response.status}]`,
        status: response.status,
        position: -1,
        query,
        type: Type.ERROR,
      }

      bus.trigger(BusEvent.MSG_CONNECTION_ERROR, {
        ...errorPayload,
        status: response.status,
      })
      bus.trigger(BusEvent.MSG_CONNECTION_ERROR, errorPayload)

      return await Promise.reject(errorPayload)
    }
```
2. Then, I would augment our generic error payload so that it includes the request response, to something like this:
```TypeScript
return await Promise.reject({
      status: response.status,
      error: response.statusText,
    })
```
This way, we'll have a nice explanation in the bottom:
CleanShot 2023-10-09 at 12 34 47@2x

Note: I haven't fully tested this code, just gave it a quick shot, so you'll need to experiment a little bit on your own. 
The SQL statement in the original issue does not trigger 428 in my case, I've needed to make it longer, like so:

```sql
select 
'a'||x,
'b'||x,
'c'||x,
'd'||x,
'e'||x,
'f'||x,
'i'||x,
'j'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
'g'||x,
'h'||x,
'i'||x,
'j'||x,
'k'||x,
'l'||x,
'f'||x,
's'||x,
1,2,3,4,5,6,7,8,9,10,11,12

 from long_sequence(100000);
 ```

insmac:
@jerempy, I have checked out your fix locally and it does exactly what it says. 

My only small comment would be to extract the `if (response.status === 408 || response.status >= 500)` conditional to a separate file, perhaps in the `utils` folder, where the `questdb.ts` file lies, so that if we need to display a specific state somewhere, it will be consistent.

insmac (approved):
@jerempy - All good 👍

insmac:
@jerempy all good, merging!

PR #207, opened 2023-10-13
feat(web-console): add localisation support for timestamps

### LINKED ISSUE #60: [Console] Improve i18n for notifications
Currently, notifications show time in the format `HH:MM:SS`:

2020-06-19-153902_grim

We want instead to check the browser locales and display the time according to the user location/preferences. Date formatting should be done either natively (i18n module, cf. [here](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/i18n) for documentation) or via [date-fns](https://date-fns.org/docs/Getting-Started).

msk4862:
The PR fixes #60 

Some example screenshots:

Screenshot 2023-10-13 at 7 24 14 PM

Screenshot 2023-10-13 at 7 18 19 PM

Screenshot 2023-10-13 at 7 26 30 PM


packages/web-console/src/scenes/Notifications/Notification/Timestamp/index.tsx
```diff
@@ -22,20 +22,46 @@
...
+export const Timestamp = ({ createdAt }: Props) => {
+  const userLocale = useMemo(() => {
+    return navigator.languages && navigator.languages.length
+      ? navigator.languages[0]
+      : navigator.language
+  }, [])
+
+  return (
+    <TimestampText color="gray2">
+      [{format(createdAt, "pppp", { locale: localeFromLanguage(userLocale) })}]
```
msk4862, line 64: Let me know if we want to [change format string](https://date-fns-interactive.netlify.app/) to something else

msk4862:
hey @insmac @oliver-daniel , can you please take a look at the PR and let me know in case of any suggestions?

packages/web-console/src/scenes/Notifications/Notification/Timestamp/index.tsx
oliver-daniel: Hey! The first thing that jumps to mind is that this kind of formatting might be really useful across a wide berth of contexts within `web-console`, not just here in the Logs/Notifications scene. What would you think of refactoring some of this code (i.e., the locale detection, registering known formats, etc.) into a reusable module in `utils`?

msk4862: make sense 👍

msk4862:
@oliver-daniel I've made the suggested changes, can you please take a look.

bluestreak01 (approved):

PR #212, opened 2023-10-23
chore(web-console): rename Table.name to Table.table_name

marregui:
linked to https://github.com/questdb/questdb/pull/3857 

QuestDB's `tables()` returns now `table_name` instead of `table`. This PR addresses this conundrum


marregui:
depends on https://github.com/questdb/questdb/pull/3857 to be able to pass the tests with success

PR #213, opened 2023-10-24
Fix `showTables` for Backwards Compatibility with `name` Property in QuestDB

argshook:
## PR Description

This PR includes a fix for the `showTables` function to ensure backwards compatibility with the `name` property. The upcoming versions of QuestDB will change the column name from `name` to `table_name` for the `tables()` query. 

The web console has already started using the `table_name` property, but it's crucial to maintain compatibility with older versions of QuestDB. 

Please note that this backwards compatibility code can be removed once the new version of QuestDB is released.

Commit log:

* fix `showTables` to be backwards compatible with `name` property
    Upcoming QuestdB versions will change the column name from `name` -> `table_name` for `tables()` query.
    
    The web console is already using `table_name` property, but it should be backwards compatible with older versions of QuestDB.
    
    Once the new version of QuestDB is released, we can remove this
    backwards compatible code.

### DIFF
--- a/packages/web-console/src/utils/questdb.ts
+++ b/packages/web-console/src/utils/questdb.ts
@@ -348,22 +348,37 @@ export class Client {
   }
 
   async showTables(): Promise<QueryResult<Table>> {
-    const response = await this.query<Table>("tables();")
+    type BackwardsCompatibleTable = Table & {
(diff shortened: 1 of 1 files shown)


insmac (approved):

marregui (approved):
LGTM

PR #220, opened 2023-11-08
feat(web-console): update pane splitters to real time

insmac:
The current splitter implementation is not flexible enough for the upcoming Import UI update. The new solution, Allotment, is derived from the VS Code core codebase, allowing for far bigger flexibility. 

This also fixes issues with the older splitter (i.e. jumping after drag end etc).

Screen recording:

https://github.com/questdb/ui/assets/1871646/6d30e20e-0bce-44ca-8631-e6b33af186b8


bluestreak01 (approved):
wow, the new splitters are epic! Great job @insmac !

PR #225, opened 2023-11-10
feat(web-console): add `current_user` info in the top bar

insmac:
Add user info next to the instance name (if available, not set on the screenshot below).

CleanShot 2023-11-10 at 10 21 03@2x

### DIFF
--- a/packages/web-console/src/components/TopBar/version.tsx
+++ b/packages/web-console/src/components/TopBar/version.tsx
@@ -5,6 +5,8 @@ import { Box } from "@questdb/react-components"
 import * as QuestDB from "../../utils/questdb"
 import { User as UserIcon } from "@styled-icons/remix-line"
 import { Text } from "../Text"
+import { selectors } from "../../store"
+import { useSelector } from "react-redux"
 
 type ServerDetails = {
   instance_name: string | null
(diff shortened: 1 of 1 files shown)


glasstiger:
Will this update after every SQL command?
Just asking because if the user is logged in as `my_user`, and runs something like:
```
ASSUME SERVICE ACCOUNT my_service_account;
```
then `my_user` has to change to `my_service_account`.

Basically after the above SQL the user switched to the service account, and `SELECT CURRENT_USER()` will return the new username.

glasstiger:
not sure how it is decided when to run `tables()` and update the list of tables on the left panel, but maybe this could work on a similar way.

insmac:
@glasstiger thanks for your input, good catch! 🥇 👍 

In short - no, the current version fetches all the variables in a single query (`SELECT instance_name, instance_rgb, current_user`) and renders accordingly, but this is on load only, but it can be easily modified as needed.

What I can do easily is to re-run this query once a particular type of query is run in the editor, i.e. we refresh `tables()` for DDL queries, this can be resfreshed as well.

glasstiger:
"What I can do easily is to re-run this query once a particular type of query is run in the editor, i.e. we refresh tables() for DDL queries, this can be refreshed as well." -> as discussed, this seems to be a good workaround for now.

Later it is probably worth adding  a new type: `DCL`, which would trigger access control related refreshes, such as current user or list of permissions for the current user.

glasstiger (approved):

PR #229, opened 2023-11-13
chore(web-console): add cloud Import disclaimer texts

insmac:
Adjust import panel disclaimer texts based on an instance type.

On prem:
CleanShot 2023-11-13 at 12 55 53@2x

Cloud:
CleanShot 2023-11-13 at 12 55 42@2x


packages/web-console/src/utils/isCloud.ts
```diff
@@ -0,0 +1,3 @@
+export const isCloud = () => {
+  return /questdb\.(net|com)$/.test(window.location.hostname)
```
bluestreak01, line 2: this is a sub-optimal way to find out if copy is supported, i will port 

```sql
show server_conf
```

to OSS and web console could call that instead

bluestreak01 (changes requested):

packages/web-console/src/utils/isCloud.ts
insmac: Very good, just ping me when this is available and I'll refactor the code 👍

packages/web-console/src/scenes/Import/ImportCSVFiles/upload.tsx
```diff
@@ -80,21 +93,22 @@ export const Upload = ({ files, onFilesDropped, dialogOpen }: Props) => {
                 name and try again.
               </Text>
             )}
+            <Info>
+              {isCloud() ? (
+                <InfoText>
+                  Note: <CopySQLLink /> is not available in the QuestDB Cloud
```
oliver-daniel, line 99:
```suggestion
                  Note: <CopySQLLink /> is not available in QuestDB Cloud
```

packages/web-console/src/utils/isCloud.ts
oliver-daniel: Implemented as part of `show parameters`.

oliver-daniel: Closest I could get to an appropriate fix looks something like:

```js
import { useContext, useEffect, useState } from 'react'
import { QuestContext } from "../providers"

type Result = {
  'property_path': string
  'env_var_name': string
  'value': string | null
  'value_source': string
  'sensitive': string
  'dynamic': 'false'
}

export const isCopyEnabled = () => {
  const { quest } = useContext(QuestContext)
  const [value, setValue] = useState(false)
  useEffect(() => {
    void quest.query<Result>(`(show parameters) where property_path ilike 'cairo.sql.copy.root'`)
    .then((result => {
      if (result.type === 'dql' && result.count > 0) {
        console.log(result.data)
        setValue(result.data[0].value !== null)
      }
    }))
  }, [quest])

  return value
}
```

No idea if this will actually be helpful for you @insmac, but I'd like to get more knowledgeable on how to actually work with the parts of the web console that interact with the backend.

oliver-daniel: or, like, idk,

```sql
with
data as (show parameters)
select value = null as is_disabled from data
where
  property_path ilike 'cairo.sql.copy.root'
```

bluestreak01 (approved):

PR #231, opened 2023-11-13
fix(web-console): fix query execution when line comments are present

insmac:


puzpuzpuz:
The following input doesn't seem to be handled correctly when the cursor at the comment on the first line:
```
select 1, '' -- test
select 2;
```

Screenshot from 2023-11-15 18-32-09

insmac:
> The following input doesn't seem to be handled correctly when the cursor at the comment on the first line:
> 
> ```
> select 1, '' -- test
> select 2;
> ```
> 
> Screenshot from 2023-11-15 18-32-09

This appears to be an overall issue with splitting queries. Our client-side parser treats this query as one, sending both SELECTs to the server. We can't assume that line comments can be a query delimiter, because of statements like these:

```
SELECT
col1,
col2, -- do not select this
col3
FROM table;
```

Overall - doesn't seem to be a quick fix.

puzpuzpuz:
Let me approve the PR then. Could you create a GH to track that corner case?

insmac:
@puzpuzpuz Will create a GH issue for the above SQL, thanks

PR #232, opened 2023-11-14
Add 'Return to Cloud' Button

argshook:
This PR includes the addition of a 'Return' button.

* show `back` button when `returnTo` URL query param is present
* the back button links to the URL specified in `returnTo` query param
* on hover, the button shows `Back to the Cloud`
* optionally, `returnToLabel` URL param can be specified to override the default label
* once page is loaded, both `returnTo` and `returnToLabel` are removed from the URL
* values are saved in indexedDB to persist across page reloads
* upon new values of `returnTo` and `returnToLabel` being set, the old values are overwritten in indexedDB

# screenshots

| before | after |
| --- | --- |
| image | image |
| | when hovered: image |

---

This PR includes the following changes:

- Cleaned up the `Menu` component by removing unused declarations, imports, and state.
- Added a new `BackButton` component.
- Fixed `PopperHover` modifiers spread.
- Made space for the `BackButton` in the UI.
- Set up indexedDB to store `returnTo` and `returnToLabel`.
- Read URL params, stored them to indexedDB, and removed them from the URL.
- Used `<a>` for linking.
- Conditionally displayed the return button in the topbar.
- Fixed a visual bug in `IconWithTooltip` modifier.
- Fixed a typo in the `align` prop.
- Used `PrimaryToggleButton` in `BackButton` for consistent styling.

<details><summary>Full commit log</summary>

* clean `Menu` component
    * remove unused declarations
    * remove unused imports
    * remove unused react state
* add `BackButton` component
* fix `PopperHover` modifiers spread
    when modifiers are spread first, they get overriden by default ones.
* make space for `BackButton`
* setup indexedDB to store `returnTo` and `returnToLabel`
* read URL params, store to indexedDB and remove from URL
* use `<a>` for linking
* conditionally display return button in the topbar
* fix IconWithTooltip modifier
    IconWithTooltip had `offset` modifier set, but it did not work, due to
    inability to override defaults.
    
    in commit adb6df41 the `modifiers` spread was
    fixed to allow overriding default modifiers (like `offset`, `arrow` and
    `eventListeners`).
    
    This made `offset` modifier set in `IconWithTooltip` to start working.
    But since it didn't work before, the change in commit adb6df41 unearthed
    a visual bug.
    
    thus, this commit removes the `offset` modifier set in `IconWithTooltip`
    to fix the visual bug.
* fix typo in `align` prop
* use `PrimaryToggleButton` in `BackButton`
    this way the style is consistent with other buttons

</details>


insmac (approved):
I wonder, wouldn't it be possible to go down to 4.5 to make the hover state a square?

| Before | After |
|--------|--------|
| CleanShot 2023-11-14 at 15 59 53@2x | CleanShot 2023-11-14 at 16 00 16@2x |

bluestreak01:
i agree, the buttons should be the same size, look & feel, also glyphs should be propotionate (weight?)

bluestreak01 (changes requested):
lets update the button L&F and we're good to go :)

argshook:
Thanks for the review! I updated the button look. Please let me know if anything else should be adjusted, thanks!

insmac (approved):
Looks & works as intended

PR #235, opened 2023-11-16
feat(web-console): add `DEDUP` support in `Copy schema to clipboard`

insmac:
Closes https://github.com/questdb/ui/issues/184

### Adding dedup configuration to table schema

If [`DEDUP UPSERT KEYS(key1, key2)`](https://questdb.io/docs/concept/deduplication/#example) has been used when creating a table, it should also be included in the generated DDL statement in `Copy schema to clipboard`.

Steps to test:

1. Create a table that has deduplication enabled
```sql
CREATE TABLE TICKS (
    ts TIMESTAMP,
    tk SYMBOL,
    p DOUBLE
) TIMESTAMP(ts) PARTITION BY DAY WAL
DEDUP UPSERT KEYS(ts, tk);

INSERT INTO TICKS VALUES ('2023-07-14', 'QQQ', 78.56);
INSERT INTO TICKS VALUES ('2023-07-14', 'AAA', 78.56);
INSERT INTO TICKS VALUES ('2023-07-14', 'QQQ', 78.56); -- this is a dupe, won't be inserted
SELECT * FROM TICKS; -- should show 2 records
```

### DIFF
--- a/packages/web-console/src/components/CreateTableDialog/index.tsx
+++ b/packages/web-console/src/components/CreateTableDialog/index.tsx
@@ -31,6 +31,7 @@ export const CreateTableDialog = () => {
         column: column.name,
         type: column.type,
       })),
+      dedup: false,
     })
     appendQuery(tableSchemaQuery, { appendAt: "end" })
     dispatch(actions.query.toggleRunning())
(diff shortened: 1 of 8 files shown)


javier (approved):
it works as described!

PR #241, opened 2023-11-21
feat(web-console): enhanced SQL editor autocomplete

insmac:
Relies on https://github.com/questdb/questdb/pull/3975

This PR introduces column autocomplete and a number of improvements in the autocomplete provider.

The syntax supported for it is as follows:
```sql
SELECT <column hint|table hint> FROM table WHERE <column hint> = '';
SELECT <column hint|table hint>, <column hint> FROM table WHERE|ORDER BY|GROUP BY <column hint>
SELECT DISTINCT <column hint>
```

Column hints are context-aware: the editor will suggest columns from a specific table if `FROM` is specified in a given query. This works for both column selection and filters.

Additional enhancements:
- Autocomplete now works in multi-line queries, like so:
```sql
SELECT
<column hint within the table context if FROM specified>
FROM
<table hint>
WHERE|ORDER BY|GROUP BY
<column hint within the table context>
```
- `JOIN` is supported for table autocomplete:
```sql
SELECT <column hint> FROM <table hint> JOIN <table hint> ON <column hint> = <column hint>
```
- The completion provider method now works in the context of a query under the cursor, enhancing its precision.
- ALTER TABLE is now supported:
```sql
ALTER TABLE <table hint> ALTER COLUMN <column hint within the table context>
ALTER COLUMN <column hint>
```
- Additional supported keywords for table autocomplete:
```sql
BACKUP TABLE
DROP TABLE
REINDEX TABLE
RENAME TABLE
RENAME TABLE
TRUNCATE TABLE
VACUUM TABLE
UPDATE
```

Additional changes:
- Update Monaco Editor to the latest stable, 0.44.0
- Refactor the deprecated `deltaDecorations` into `editor.createDecorationsCollection()` method


packages/web-console/src/scenes/Editor/Monaco/questdb-sql/createSchemaCompletionProvider.ts
```diff
@@ -1,55 +1,129 @@
...
+  editor: IStandaloneCodeEditor,
+  tables: Table[] = [],
+  informationSchemaColumns: InformationSchemaColumn[] = [],
+) => {
   const completionProvider: monaco.languages.CompletionItemProvider = {
-    triggerCharacters: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz \"".split(
-      "",
-    ),
+    triggerCharacters:
+      'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz\n "'.split(""),
```
oliver-daniel, line 31: Trivial readability nit:
```suggestion
      Array.from('ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz\n "'),
```

puzpuzpuz:
> * Additional supported keywords for table autocomplete

Probably we should also add support for `show parameters`.

puzpuzpuz:
Notice one weird thing: if you start typing select, or from, or where the autocomplete shows columns/tables like it should. But if you dismiss the autocomplete and then press Ctrl + space autocomplete will show a different list of options (same as before this patch).
Screenshot from 2023-11-24 18-43-24

puzpuzpuz:
Another thing to improve: equal column names may be merged when typing SELECT. Here I have 10 identical tables and the list shows 10 duplicate sets of columns.
Screenshot from 2023-11-24 18-46-53

puzpuzpuz:
Probably I'm asking too much, but it would be great if autocomplete also worked with aliases:
```sql
select s_col, ts from x as y where y.
```

Currently, autocomplete shows no options when I type `y.`.

insmac:
> Another thing to improve: equal column names may be merged when typing SELECT. Here I have 10 identical tables and the list shows 10 duplicate sets of columns. Screenshot from 2023-11-24 18-46-53

It's definitely possible, For columns that appear in multiple tables I can add something like:
```
column_name (table1, table2, ...)
```

puzpuzpuz:
> It's definitely possible, For columns that appear in multiple tables I can add something like

That would be great.

insmac:
> > * Additional supported keywords for table autocomplete
> 
> Probably we should also add support for `show parameters`.

So I have checked and we support these:
```
SHOW COLUMNS FROM <table name>
SHOW PARTITIONS FROM <table name>
```
Looking into [the docs](https://questdb.io/docs/reference/sql/show/), I don't see any other statement that can be supported - or am I wrong?

puzpuzpuz:
> Looking into [the docs](https://questdb.io/docs/reference/sql/show/), I don't see any other statement that can be supported - or am I wrong?

It's a [recent change](https://github.com/questdb/questdb/pull/3972) that was shipped in 7.3.5. I don't think we have any documentation around it. Same with [`show server_version`](https://github.com/questdb/questdb/pull/3901).

cc @goodroot

insmac:
> > * If there are many, show the list of the tables in which the name is present, separate by a comma. This list is auto-truncated by Monaco editor with an ellipsis).
> 
> For some reason, I still see column type instead of list of tables - I have 10 tables with identical schema. I'm on [ba3d4ab](https://github.com/questdb/ui/commit/ba3d4ab80d7068212fea3eaaaf680ef2310a1c73) and made sure to rebuild React components, rerun the test server and flush browser cache.
> 
> Screenshot from 2023-11-27 14-50-07
> 
> Also, when I Ctrl + space after a SELECT, I still see the autocompletion list with <<, <<= and other functions and operators, not the one with column names.
> 
> Most likely, I'm doing something wrong.

In short - you're not seeing duplicates - the values are unique, but there is a type instead of table listing on the right?

puzpuzpuz:
> In short - you're not seeing duplicates - the values are unique, but there is a type instead of table listing on the right?

Yes, that's correct. I don't see the table list, but the type only.

insmac:
> > In short - you're not seeing duplicates - the values are unique, but there is a type instead of table listing on the right?
> 
> Yes, that's correct. I don't see the table list, but the type only.

Interesting, I have created a few tables with the same structure as on your screenshot and this is what I get:
CleanShot 2023-11-27 at 15 12 35@2x

Could you try in a different browser or incognito? And if this does not help, perhaps you could log the array here:
https://github.com/questdb/ui/pull/241/files#diff-c460c7a05460c60c5cfa78d77188acab15fd62afb1d35eda1c6f910e5a38a2ddR30

The logic is simple - if that array contains a single item (single table), column type should be displayed, otherwise it will map out a list of table names.

puzpuzpuz:
> Could you try in a different browser or incognito?

Tried incognito mode and the table lists were there. Maybe this has something to do with editor history?

In any case, let's consider the following things to be resolved:
> For some reason, I still see column type instead of list of tables - I have 10 tables with identical schema.

> Also, when I Ctrl + space after a SELECT, I still see the autocompletion list with <<, <<= and other functions and operators, not the one with column names.

bluestreak01:
I have `cpu` table with columns `a`, `b`. I think we should not be prompting tables names where columns go (is that feasible to do):

image

insmac:
@bluestreak01 

> I think we should not be prompting tables names where columns go (is that feasible to do).

Do you mean that we shouldn't display the table name in brackets (next to the column name) or actual table names in the above list?

> Also do you think we can add intelligence to narrow columns down using column choices already made?
I thought about implementing it, but it's not a trivial thing to do. We would need to know which columns are already used for a given query, which is a complex thing to do with regexp, less so with a proper AST Tree, but we don't have that yet.

Grafana (among others) doesn't have this capability as well:
CleanShot 2023-12-14 at 19 58 56@2x

it's definitely doable, but it is a fairly complex task.

bluestreak01:
all good @insmac i was just curious if it is a good idea. On a second thought what you're showing now is better. If user wants to use table name for column name who are we to stop them :)

PR #248, opened 2023-12-01
feat(web-console): Add URL deep-linking for UI panels

### LINKED ISSUE #245: Route directly to CSV Import flow from URL
As a user, I can be routed directly to the Web Console in such a way that the CSV Import panel is immediately opened and ready for use.

insmac:
Add URL deep-linking for the UI panels:
```
?topPanel=p1 (default: p1 - tables - this is future-proofing, as there will be more in near future)
?bottomPanel=p0|p1|p2 (default: p0)
?sidebar=p1|p2 (default: none)
```

Panel alignment screenshot:
web-console-panels

EDIT: Closes #245


bluestreak01:
We could assign panels and their states integer IDs, this way we can keep backward compatibility even if UI changes. Dependency on panel human readable names is not ideal.

bluestreak01 (changes requested):
if you have less that 3 tables to have zero-state panel visible, go to this URL: http://localhost:9999/?bottomPanel=result execute a query and observe query result not being displayed

goodroot:
Tidying up 🧹 

PR #250, opened 2023-12-06
chore(web-console): replace jQuery event bus with eventemitter3

insmac:
This PR ditches jQuery event bus used for cross-framework communication and introduces a lightweight abstraction based on [EventEmitter3](https://www.npmjs.com/package/eventemitter3), keeping the features intact.

Charts file has been converted to TypeScript quickly, just so that it can support the new bus system. It does not have a proper type system though, as it's very old anyway and being the very last piece of code still relying on jQuery, going to be discarded fairly shortly.


bluestreak01 (changes requested):
I get error visuals on successful query execution:

image

insmac:
@bluestreak01 fixed

bluestreak01 (approved):

PR #253, opened 2023-12-11
chore(web-console): display result zero state at all times

insmac:
There has been user feedback that finding certain action triggers, like CSV Import, proved to be difficult after the UI refresh.

This change makes the "new user" zero state visible at all times when the Web Console is launched, apart when running in read-only mode.

CleanShot 2023-12-11 at 18 28 52@2x


bluestreak01:
Lets wait with this please. It has been single real-world user feedback out of thouthands deployed instances. Our CSV upload is a source of frustration for many, making zero state suggestive of CSV could be handy after we released updated CSV import.

While we're on this PR, is that possible to make upload icon the same?

image

insmac:
Closed in favour of https://github.com/questdb/ui/pull/258

PR #263, opened 2024-01-09
chore(web-console): update release info API call code

insmac:
- Use the AWS Lambda endpoint to fetch the latest QuestDB OSS release info instead of calling Github API directly.
- Make sure the error is caught when the call cannot be made.
- The same for News API.

### DIFF
--- a/packages/web-console/src/utils/questdb.ts
+++ b/packages/web-console/src/utils/questdb.ts
@@ -504,12 +504,11 @@ export class Client {
   async getLatestRelease() {
     try {
       const response: Response = await fetch(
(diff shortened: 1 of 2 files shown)


packages/web-console/src/utils/questdb.ts
```diff
@@ -504,12 +504,11 @@ export class Client {
   async getLatestRelease() {
     try {
       const response: Response = await fetch(
-        `https://api.github.com/repos/questdb/questdb/releases/latest`,
+        `https://github-api.questdb.io/github/latest`,
```
bluestreak01, line 507: this looks like incorrect URL?

bluestreak01 (changes requested):

PR #272, opened 2024-03-27
feat(web-console): implement authorization logic

insmac:


packages/web-console/src/providers/SettingsProvider/types.ts
```diff
@@ -0,0 +1,30 @@
...
+  "acl.oidc.client.id": string
+  "acl.oidc.host": string
+  "acl.oidc.port": number
+  "acl.oidc.tls.enabled": boolean
+  "acl.oidc.authorization.endpoint": string
+  "acl.oidc.token.endpoint": string
+  "acl.oidc.pkce.required": boolean
+  "acl.basic.auth.realm.enabled"?: boolean
+  "questdb.type"?: "OSS" | "EE"
+  "questdb.version"?: string
```
glasstiger, line 12: why are questdb.type and questdb.version optional?
i would understand if the acl fields are made optional, they are only present in EE.

packages/web-console/src/modules/OAuth2/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+      },
+      body: new URLSearchParams(payload),
+    },
+  )
+}
+
+export const hasNoAuth = (config: Settings) => config["questdb.type"] === "OSS"
+
+export const hasUIAuth = (config: Settings) =>
+  !hasNoAuth(config) && (config["acl.oidc.enabled"] || !config["acl.basic.auth.realm.enabled"])
```
glasstiger, line 76: i think this should be:
```
export const hasUIAuth = (config: Settings) =>
  config["acl.enabled"] && !config["acl.basic.auth.realm.enabled"]
```

The problem with the above is that if we disable auth in EE (acl.enabled = false) on server side, then this could still evaluate to true.
I need to expose `acl.enabled` property to the UI.
Was not keen on this, but I think we have to.

glasstiger (approved):

PR #274, opened 2024-04-02
fix(web-console): combine buffer init into a single hook

ngnic:
Querying the buffer store through Dexie's live query caused frequent re-renders on any component that used the EditorProvider Context as each character entered into the buffer would trigger a change in state.

The flickering borders indicate components that are re-rendered. 

Before:

https://github.com/questdb/ui/assets/7108642/c4da847e-d81d-47f1-b32b-2658e9a8b26d

After:

https://github.com/questdb/ui/assets/7108642/025477f7-8e31-44b3-8b13-6e5e852663d7

The buffer is still being persisted correctly into indexedDB.

https://github.com/questdb/ui/assets/7108642/4130f759-3719-4a2a-8904-9960e8373c61


insmac (approved):
Checked the fix in profiler + tested the persistence layer.

PR #281, opened 2024-05-06
feat(web-console): create table dialog supports VARCHAR

jerrinot:
The console uses it as a default column type instead of STRING. It's also possible to define a VARCHAR column when import a CSV file.


bluestreak01:
when executing SQL the OSS web console sends cryptic bearer token:

image

@insmac

insmac (approved):
LGTM, both Create Table and CSV Import work.

PR #284, opened 2024-05-08
chore(web-console): add posthog support

ngnic:
This PR adds support for posthog. Posthog is set to disabled by default. Settings are fetched from QDB.  Have attached a screenshot to show that no calls are made to posthog if it's disabled.
Screenshot 2024-05-08 at 8 36 48 PM


packages/web-console/src/index.tsx
```diff
@@ -68,17 +69,19 @@ ReactDOM.render(
...
-              <GlobalStyle />
-              {ReactDOM.createPortal(<ToastContainer />, document.body)}
-              <LocalStorageProvider>
-                <FadeSlow />
-                <FadeReg />
-                <Layout />
-              </LocalStorageProvider>
-            </QuestProvider>
-          </AuthProvider>
+          <PosthogProviderWrapper>
```
insmac, line 72: I'm starting to think we should wrap the app with `PostHogProvider` only when it's enabled in settings, so the code does not parse the provider if `posthog.api.key` is `null`. Not sure if this impacts app start time, but it's a hunch I have

ngnic: https://github.com/PostHog/posthog-js/blob/main/react/src/context/PostHogProvider.tsx 

The client is only initialised if the api key is provided. If the app is only conditionally wrapped, there will need to be additional checks if the client is undefined when the context is used. There are already [checks](https://github.com/PostHog/posthog-js/blob/108d8f88cf39926aa8d595fb39d14ddb7e4e8085/src/posthog-core.ts#L725-L727) in place within the client itself to ensure that the functions are no-ops if the client isn't initialised.

insmac: OK, got it. Makes sense, I see they check for it

insmac (approved):
LGTM. Checked build size, it's not inflated by the SDK addition in any way.

insmac (approved):

PR #291, opened 2024-05-15
feat(web-console): system info - suspension and configuration

insmac:
This PR depends on https://github.com/questdb/questdb/pull/4607 and https://github.com/questdb/questdb-enterprise/pull/439

WAL tables can be suspended for writing for multiple reasons. This PR adds the indication and restart function into the UI, as well as other things in the Tables UI widget. 
In addition to showing the suspension details, the popover contains `Restart transaction` action which could either be performed from the first applicable txn, or a specified number (optional)

Side changes:
- Add filtering via keyword input - search in both table and column names, highlight the found portion of the text.
- Simplify the Table components, remove regacy rxjs methods, use the reducer hook to handle view state
- Store `information_schema.columns()` result in Redux, since it's shared by both Tables UI and SQL Editor, adjust the Editor's completion provider code to use redux selector.

## Tables - changes to the header action bar
Screenshot 2024-06-14 at 10 30 32

- A few people have reported not being able to find "create table" icon easily, therefore a "plus" one has been added in the top right. 
- Table schema is now refreshed automatically in almost all needed situations, including back to browser tab focus, so there is virtually no need to have the "refresh" action that prominent. There is a dropdown "action menu" added instead, which now houses both refresh and copy all schemas items.

## Tables - list view with a suspended table
initial-view

## Tables - details of the suspended table as a popover
error-expanded

## Filter by keyword
filter

Test changes:
- Added two CY commands for working with Web Console, `getByDataHook()` and `interceptQuery(query, response)`, which are both self-explanatory.
- Changed the Cypress config timeout so that spurious CI timeouts (which are not related to the code and never happen locally) don't pop up randomly, depending on the CI load.


glasstiger:
Testing
---------
1. Check `wal_tables()` output, there should not be any suspended tables.
2. Run `alter table trades suspend wal with 'DISK FULL', 'test error'`, check `wal_tables()' output.
3. `trades` should be suspended and error tag and message should be populated.
Suspension also should be indicated in the table view next to the table, and when opening the suspension details panel WAL apply can be resumed by clicking on `Resume WAL`.

glasstiger:
Testing WAL suspension:
-------------------------
1. Check `wal_tables()` output, there should not be any suspended tables.
2. Run `alter table trades suspend wal with 'DISK FULL', 'test error'`, then check `wal_tables()' output.
3. At this point `trades` table should be suspended and error tag and message should be populated as `DISK FULL` and `test error`.
Suspension also should be indicated in the table view on the left side of the screen.
When opening the suspension details panel there should be more details displayed, including a link to a page in the docs relevant for the issue.
WAL apply can also be resumed by clicking on `Resume WAL`, or with the `alter table trades resume wal` SQL command.

Valid error tags are: "DISK FULL", "TOO MANY OPEN FILES", "OUT OF MMAP AREAS" and "OUT OF MEMORY"

glasstiger:
Testing server configuration warnings:
--------------------------------------

Testing server configuration warnings with simulation:
------------------------------------------------------
1. Add `cairo.enable.warnings.simulation=true` to your server config.
2. When loading the Web Console there should not be any warnings.
3. Run `select simulate_warnings('UNSUPPORTED FILE SYSTEM', 'Unsupported file system [dir=/questdb/path/dbRoot, magic=0x6400A468]')`. This should return with `true`. If not, warning simulation is not enabled, go back to step 1, and check if you added the config to server.conf.
4. A warning should appear with the above message after reloading the console.
5. Run `select simulate_warnings('TOO MANY OPEN FILES', 'fs.file-max limit is too low [current=1024, recommended=1048576]')`. This should also return with `true`.
6. Reload the console again. Now you should see 2 warnings.
7. Run `select simulate_warnings('', '')`. This should also return with `true`.
8. Reload the console again. There should not be any warnings.
 
All warnings should include a link to a page in the docs relevant for the problem.

Valid warning tags are: "UNSUPPORTED FILE SYSTEM", "TOO MANY OPEN FILES" and "OUT OF MMAP AREAS"

glasstiger (approved):

PR #300, opened 2024-06-17
feat(web-console): show login dialog for QuestDB OSS when acl is enabled

jerrinot:
This shows a login form when QuestDB has HTTP auth enabled. 

depends on https://github.com/questdb/questdb/pull/4672

### DIFF
--- a/packages/web-console/src/modules/OAuth2/views/login.tsx
+++ b/packages/web-console/src/modules/OAuth2/views/login.tsx
@@ -203,8 +203,9 @@ export const Login = ({
   onBasicAuthSuccess: () => void
 }) => {
   const { settings } = useSettings()
+  const isEE = settings["release.type"] === "EE"
   const [errorMessage, setErrorMessage] = React.useState<string | undefined>()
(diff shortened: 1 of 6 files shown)


packages/web-console/src/modules/OAuth2/views/login.tsx
```diff
@@ -203,21 +203,35 @@ export const Login = ({
   onBasicAuthSuccess: () => void
 }) => {
   const { settings } = useSettings()
+  const isEE = settings["release.type"] === "EE"
   const [errorMessage, setErrorMessage] = React.useState<string | undefined>()
+
+  const httpBasicAuthStrategy = isEE ? {
+    query: "alter user 'admin' create token type rest with ttl '1d' refresh transient",
```
glasstiger, line 210: username is hardcoded as 'admin', should be the user currently logged in

glasstiger: fixed it

glasstiger (approved):

PR #308, opened 2024-07-16
chore(browser-tests): Refactor waiting for editor ready state

insmac:
This PR is aimed to further reduce the possible flakiness on CI, which, although rare, sometimes manifests itself when the tests start to run before the editor is fully ready UI wise.

**Changes:**
- Introduced the `window.editor_ready` global flag, which is set at the very end of the editor initialization process, when all the completion providers finish setup. This is in contrast to checking for mere `textarea` presence, which can happen much earlier.
- Added `waitForEditorLoad` Cypress command to use in `before` and `beforeEach` hooks.
- Converted "service" queries (DDLs, warning simulation) to `cy.request()` instead of typing them manually in the editor - there is zero user value in the latter, and it significantly slows down the text execution time, as they are usually quite long.
- Configured a slightly longer keystroke delay for typing inside the editor to prevent occasional test hiccups (very rare, but still) when using multi line queries, various cursor keys, etc.

### DIFF
--- a/packages/browser-tests/cypress/commands.js
+++ b/packages/browser-tests/cypress/commands.js
@@ -82,7 +82,9 @@ beforeEach(() => {
 
   cy.request({
     method: "GET",
-    url: `${baseUrl}/exec?query=${encodeURIComponent(`select simulate_warnings('', '');`)}`,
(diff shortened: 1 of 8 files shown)


goodroot:
Tidying up 🧹 

PR #309, opened 2024-07-17
chore(web-console): Small UI fixes related to Tables

insmac:
- Remove the `+` icon in the Tables header - it will come back when the Create Table view is opened in the left side of the screen.
- Fix for clicking the disabled `Create Table` icon in read-only mode - it should not be possible to show it and clicking the icon should be a no-op.
- Fix table filter: it should show only the tables with a match and highlight.

### DIFF
--- a/packages/browser-tests/cypress/integration/console/schema.spec.js
+++ b/packages/browser-tests/cypress/integration/console/schema.spec.js
@@ -130,3 +130,24 @@ describe("questdb schema with suspended tables with Linux OS error codes", () =>
     });
   });
 });
+
+describe.only("questdb schema in read-only mode", () => {
+  before(() => {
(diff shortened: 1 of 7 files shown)


bluestreak01 (changes requested):
could you add a test to ensure disabled button cannot be toggled?

javier:
image

I can select now table columns, but no table names. Since table names are shown in an autocompletion drop down right after typing `FROM` or `JOIN` this is not as bad, but it would be nicer if we have it, as some people might be used to just type the table name and enter to get the contents without writing `SELECT * FROM`

javier:
Restested after last commit and copying functionality works fine now. Thanks!

bluestreak01 (approved):

PR #318, opened 2024-07-26
chore(web-console): disable telemetry on CI runs

insmac:


insmac:
Closing, as there is no need to manually set the flag, since we are intercepting the telemetry requests in the global config.

PR #327, opened 2024-08-29
fix(web-console): fix suspension lag count

insmac:
When the WAL table is suspended, user can click on the `Suspended` button next to the table name to diagnose the problem. The current number of transactions behind is derived from `writerLagTxnCount` property, which gives incorrect results. This PR fixes it by switching to the difference between sequencer and writer txn counts.

### DIFF
--- a/packages/web-console/src/scenes/Schema/SuspensionDialog/index.tsx
+++ b/packages/web-console/src/scenes/Schema/SuspensionDialog/index.tsx
@@ -89,6 +89,9 @@ export const SuspensionDialog = ({
   const [isSubmitted, setIsSubmitted] = useState(false)
   const [error, setError] = useState<string | undefined>()
 
+  const txnLag =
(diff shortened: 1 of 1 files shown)


packages/web-console/src/scenes/Schema/SuspensionDialog/index.tsx
```diff
@@ -89,6 +89,9 @@ export const SuspensionDialog = ({
   const [isSubmitted, setIsSubmitted] = useState(false)
   const [error, setError] = useState<string | undefined>()
 
+  const txnLag =
+    parseFloat(walTableData.sequencerTxn) - parseFloat(walTableData.writerTxn)
```
glasstiger, line 93: nit: `parseInt()` might have been better, txn numbers are integers.

glasstiger (approved):

PR #328, opened 2024-08-30
feat(web-console): Fix the response to handle all case of table name

AayushSaini101:
Resolves: #324 
* Allow the web console to handle all cases of table names.
* Example: Table Name student and Student must be treated the same.

### DIFF
--- a/packages/browser-tests/cypress/integration/console/schema.spec.js
+++ b/packages/browser-tests/cypress/integration/console/schema.spec.js
@@ -19,12 +19,6 @@ describe("questdb schema with working tables", () => {
     cy.refreshSchema();
   });
   
-  it("should throw a warning when table is present with different case",()=>{
-    cy.get('input[name="table_filter"]').type("Btc_trades").should("contains","Table name must be unique")
-    cy.get('input[name="table_filter"]').type("gitLog").should("contains","Table name must be unique")
-    cy.get('input[name="table_filter"]').type("GITLOG").should("contains","Table name must be unique")
-  })
-
   it("should show all the tables when there are no suspended", () => {
     tables.forEach((table) => {
       cy.getByDataHook("schema-table-title").should("contain", table);
--- a/packages/browser-tests/cypress/integration/console/schema.spec.js
+++ b/packages/browser-tests/cypress/integration/console/schema.spec.js
@@ -18,6 +18,19 @@ describe("questdb schema with working tables", () => {
     });
     cy.refreshSchema();
   });
+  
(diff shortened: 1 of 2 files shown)


AayushSaini101:
@glasstiger can you please review this ?

glasstiger:
@AayushSaini101 change looks good but would be nice to add tests.
I think easiest is to create a cypress test which replicates the issue reported.

AayushSaini101:
@glasstiger Added thanks

packages/browser-tests/cypress/integration/console/schema.spec.js
```diff
@@ -18,6 +18,19 @@ describe("questdb schema with working tables", () => {
     });
     cy.refreshSchema();
   });
+  
+  it("should throw a warning when table is present with different case",()=>{
```
glasstiger, line 22: Hey @AayushSaini101 , it looks to me that this test does not test your change.
This test goes to the backend to create the table.
What you would need to do is use cypress to open the table schema dialog, type in "Btc_Trades", and assert that the error message appeared.

I do not think you need to create any tables on the backend.
It looks to me that the 4 tables are created before every test in a `before()` hook, so they should exist already.
No need for this code:
```
    tables.forEach((table) => {
      cy.createTable(table);
    });
```

Check out this test in this file at line 43:
```
it("should filter the table with input field", () => {
  ...
}
```

Not exactly what you need, but it is a good example how to find elements on the UI with cypress, and how to simulate typing in text into an input, for example.

AayushSaini101: Hope this works thanks @glasstiger 
```
  it("should throw a warning when table is present with different case",()=>{
    cy.get('input[name="table_filter"]').type("Btc_trades").should("contains","Table name must be unique")
    cy.get('input[name="table_filter"]').type("gitLog").should("contains","Table name must be unique")
    cy.get('input[name="table_filter"]').type("GITLOG").should("contains","Table name must be unique")
  })
```

glasstiger: Hi @AayushSaini101 , is this test passing for you when you run it?

glasstiger: Please, remove the test.
We can merge it.
It is a tiny change.

AayushSaini101: Done @glasstiger Please merge

glasstiger (approved):

PR #329, opened 2024-09-04
feat(web-console): editor tabs

insmac:
This PR introduces tabs in SQL editor:
- each tab holds editor content, cursor and scroll position in memory and in IndexedDB, so the arrangement is persisted between browser sessions.
- at least one tab has to be present at all times (this is also the initial state when the editor is empty).
- tabs behave much like those in google chrome, both functionally and visually.
- tabs can be reordered as needed.

Screen:
Screenshot 2024-09-05 at 13 03 29


javier (approved):
tabs working fine. Tested creating, reorder, history, reloading... all good!

PR #332, opened 2024-09-20
fix(web-console): Display `Copy` for suspended tables

insmac:
Display the `Copy` button for suspended tables too - it's currently not visible.


javier (approved):
Looking good!

PR #355, opened 2024-11-15
feat(ui): support for ID token

glasstiger:


insmac (approved):
Confirmed working with the provided credentials.

The only important thing is that the group with the right UUID alias has to be created upfront for EntraID to work.

PR #357, opened 2024-11-25
chore(ui): fix token refresh

glasstiger:
- no need to send code verifier in token refresh request
- PingFederate loses id token on token refresh, removed checks for id token


insmac (approved):
Tested against the `ia_azure_ropc` enterprise branch.

- Login works as previously
- Token refresh works as well

PR #371, opened 2025-01-08
chore(ui): webpack dev server proxy to take context path from env variable

glasstiger:
When a context path is configured on server side, we can set an env var for dev server to set the same context path to be used by the proxy.

For example:
```
CONTEXT_PATH=/instance2
export CONTEXT_PATH
yarn workspace @questdb/web-console start
```

To get rid of the context path:
```
unset CONTEXT_PATH
yarn workspace @questdb/web-console start
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 This repository is a monorepo hosting the implementation of QuestDB user
 interface and surrounding tooling.
 
-Currently there are these packages:
+Currently, there are these packages:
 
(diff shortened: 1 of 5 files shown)


insmac (approved):
Let's add info about CONTEXT_PATH to README.md

glasstiger:
> Let's add info about CONTEXT_PATH to README.md

updated doc.

insmac (approved):

PR #386, opened 2025-01-30
chore(web-console): reformat code using project Prettier config

insmac:
Webstorm's built in Actions on save seem to supersede the built-in Prettier config definitions: If `Reformat on save` is enabled, it will reformat according to it's own definitions, rather than relying on the `prettier.config.js`. This is a different behavior than in VSCode, when the config file is taken into account on every file save.

This PR does a full formatting pass on Web Console source code.


goodroot:
Tidying up 🧹

PR #388, opened 2025-01-31
chore(web-console): dispose inactive models on tab change

insmac:
Current logic does not dispose of inactive editor models while switching tabs, meaning the model array grows exponentially in size with every tab change/drag/creation. This is true even if the editor itself is unmounted (i.e. when switch to metric tabs), because the models reside in an array within a global window object.

This ensures the only model is the active one.


goodroot:
Rolled into https://github.com/questdb/ui/pull/406

PR #389, opened 2025-01-31
chore(web-console): reduce chart re-rendering

insmac:
The current version of Metrics re-renders the charts more than it's needed due to how the method to generate the uPlot configuration object is executed. This PR ensures each chart widget is rendered one time per refresh cycle. This can be verified by putting on a logger in the [`onCreate` method of the uPlot component](https://github.com/questdb/ui/blob/main/packages/web-console/src/scenes/Editor/Metrics/graph.tsx#L236).

Changes:
- Convert `useGraphOptions` to a method that is no longer a React hook, since there is no need to get the theme context in there anymore. 
- Refresh the uPlot configuration object only when the data is changed (= per refresh cycle).


goodroot:
Rolled into https://github.com/questdb/ui/pull/406

PR #390, opened 2025-02-03
feat(ui): support for OAuth2 state parameter

glasstiger:
If the `state` parameter is enabled in settings (`acl.oidc.state.required=true`), the Web Console generates a state value, and sends it to the Authorization Server while requesting authorization code.
The `state` is expected to be sent back to the Web Console together with the authorization code, the Web Console checks if the value of the `state` parameter is the same.
If not, it sends the user to the logout screen.

### DIFF
--- a/packages/web-console/src/providers/AuthProvider.tsx
+++ b/packages/web-console/src/providers/AuthProvider.tsx
@@ -48,7 +48,7 @@ const initialState: { view: View; errorMessage?: string } = {
   view: View.loading,
 }
 
-const defaultValues = {
+const defaultValues: ContextProps = {
   sessionData: undefined,
   logout: () => {},
   refreshAuthToken: async () => ({} as AuthPayload),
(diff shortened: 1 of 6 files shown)


packages/web-console/src/providers/AuthProvider.tsx
```diff
@@ -187,14 +188,26 @@ export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
       } else {
         // User has just been redirected back from the OAuth2 provider and has the code
         if (code !== null) {
+          const state = getValue(StoreKey.OAUTH_STATE)
+          if (state) {
```
bluestreak01, line 192: what is the logic if there is no "state" in local storage?

bluestreak01: ok, thanks

bluestreak01 (approved):

PR #393, opened 2025-02-05
feat(web-console): add button to copy query plans and result sets to clipboard in markdown format

nwoolmer:
When developing QuestDB, or helping users to debug issues, lots of screenshots of queries, schemas etc. get thrown around.

The schema issue is resolved by `SHOW CREATE TABLE tbl;`, which gives a formatted DDL.

However, due to the way the grid element works for our result sets, it is non-trivial to copy query plans to text - it is broken up, or includes `&nbsp;`.

Likewise, it is useful to copy small result sets to markdown table format, when writing documentation, or trying to share results with others.

image

Examples:

**Query Plan**

```sql
EXPLAIN
SELECT * FROM cities
PIVOT (
    SUM(population)
    FOR
        year IN (2000, 2010, 2020)
    GROUP BY country
);
```

```
Async Group By workers: 16
  keys: [country]
  values: [sum(case([population,null,year])),sum(case([population,null,year])),sum(case([population,null,year]))]
  filter: null
    PageFrame
        Row forward scan
        Frame forward scan on: cities
```

**Result Set**

```sql

SELECT * FROM 
    (SELECT * FROM trades WHERE symbol IN 'ETH-USDT')
PIVOT (
    SUM(price)
    FOR "symbol" IN ('ETH-USDT')
        side in ('buy', 'sell')
    GROUP BY timestamp
);
```

| timestamp                   | ETH-USDT_buy | ETH-USDT_sell |
| --------------------------- | ------------ | ------------- |
| 2024-12-19T08:10:00.700999Z | null         | 3678.25       |
| 2024-12-19T08:10:00.736000Z | null         | 3678.25       |
| 2024-12-19T08:10:00.759000Z | null         | 3678          |
| 2024-12-19T08:10:00.772999Z | null         | 3678          |
| 2024-12-19T08:10:00.887000Z | 3678.01      | null          |
| 2024-12-19T08:10:00.950000Z | null         | 3678          |

Changelist:

- [x] Add button to copy results to markdown
- [x] Support exporting a formatted query plan
- [x] Support exporting an actual result set as a markdown table

### DIFF
--- a/packages/browser-tests/cypress.config.js
+++ b/packages/browser-tests/cypress.config.js
@@ -13,6 +13,7 @@ module.exports = defineConfig({
     screenshotOnRunFailure: false,
     video: false,
     baseUrl: baseUrl,
+    chromeWebSecurity: false, //if it is true, cypress does not allow redirects
     viewportWidth: 1280,
     viewportHeight: 720,
     specPattern: "cypress/integration/**/*.spec.js",
--- a/packages/browser-tests/cypress/commands.js
+++ b/packages/browser-tests/cypress/commands.js
@@ -265,6 +265,24 @@ Cypress.Commands.add("loadConsoleWithAuth", (clearWarnings) => {
   }
 });
 
+Cypress.Commands.add("loadConsoleAsAdminAndCreateSSOGroup", (group, externalGroup = undefined) => {
+  cy.loadConsoleWithAuth(true);
+  cy.executeSQL(`CREATE GROUP ${group} WITH EXTERNAL ALIAS ${externalGroup || group};`);
+  cy.executeSQL(`GRANT HTTP TO ${group};`);
+  cy.logout();
+});
+
+Cypress.Commands.add("logout", () => {
+  cy.getByDataHook("button-logout").click();
+  cy.getByDataHook("auth-login").should("be.visible");
+});
+
+Cypress.Commands.add("executeSQL", (sql) => {
+  cy.clearEditor();
+  cy.typeQuery(sql);
+  cy.clickRun();
+});
+
 Cypress.Commands.add("refreshSchema", () => {
   // toggle between auto-refresh modes to trigger a schema refresh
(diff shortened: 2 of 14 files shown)


packages/web-console/src/scenes/Result/index.tsx
```diff
@@ -178,6 +179,20 @@ const Result = ({ viewMode }: { viewMode: ResultViewMode }) => {
   }, [activeSidebar])
 
   const gridActions = [
+    {
```
goodroot, line 182: For style, not mind-blowing but...

How about a green checkmark?

https://github.com/user-attachments/assets/2763c7f3-a7a9-4ece-a638-951f698285ad

To do so:

```typescript
    {
      tooltipText: "Copy result to Markdown",
      trigger: (
        <PrimaryToggleButton
          selected={isCopied}
          onClick={() => {
            navigator.clipboard.writeText(gridRef?.current?.getResultAsMarkdown() as string)
              .then(() => {
                setIsCopied(true)
                setTimeout(() => setIsCopied(false), 1000)
              })
          }}
        >
          {isCopied ? <Check size="18px" /> : <Markdown size="18px" />}
        </PrimaryToggleButton>
      ),
    },
```

Need to import Check:

```typescript
import { Check } from "@styled-icons/bootstrap/Check"
```

Then just slap this somewhere:

```typescript
const [isCopied, setIsCopied] = useState(false)
```

goodroot (commented):
This will be very handy 💥

goodroot:
One question and one small issue:

1. How many results do we expect to copy? On a 10,000 row set, I successfully copy 63.

2. When tooltip is present, displaying `Copied!` goes tooltip start and breaks "connector":


https://github.com/user-attachments/assets/ab18e1f3-e44a-4f52-b6e8-606c869a8047

goodroot (approved):
👨‍🍳

PR #405, opened 2025-03-13
feat(web-console): add materialized views support

emrberk:
Screenshot 2025-03-20 at 18 04 37

- Optimizes table listing using `GroupedVirtuoso` from `react-virtuoso`. Reduces render time in case we have 1000s of tables / mat. views.
- Adds SQL grammar support for materialized views.
- Materialized views shows the base table, and copying schema invokes `show create materialized view...` instead of `show create table`.
- Resume WAL button, in case of suspension, is moved to right-click context menu.
- Copy Schema action is available in the context menu.
- Fixes hide/show tables wording. We were showing "hide" even when it was hidden. Changes "tables" wording to "data sources" (highly subject to a discussion).
- Adds icons for different column types.
- Moves column type text to the left, changes the color.
- Replaces `react-contextmenu` with `@radix-ui/react-context-menu` due to bugs.
- Adds e2e tests for UI changes.


.github/workflows/ci.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           QDB_DEV_MODE_ENABLED: "true"
           QDB_HTTP_USER: "admin"
           QDB_HTTP_PASSWORD: "quest"
+          QDB_CAIRO_MAT_VIEW_ENABLED: "true"
```
emrberk, line 68: When materialized views is released, the variable will be true by default.
We can remove this, or leave it.

.github/workflows/tests_with_context_path.yml
```diff
@@ -68,6 +68,7 @@ jobs:
           QDB_DEV_MODE_ENABLED: "true"
           QDB_HTTP_USER: "admin"
           QDB_HTTP_PASSWORD: "quest"
+          QDB_CAIRO_MAT_VIEW_ENABLED: "true"
```
emrberk, line 71: Same stuff with https://github.com/questdb/ui/pull/405/files#r1994136742

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+  padding: 1rem;
+  font-weight: bold;
+  background: ${color("backgroundDarker")};
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  cursor: pointer;
+
+  &:hover {
+    background: ${color("selection")};
```
goodroot, line 36: Minor issue with hover shading:

https://github.com/user-attachments/assets/0515849a-9f0e-4da8-8b5a-b6b9c37c5810

* columns have uniform shade, look good
* other rows show varying length and height

Seems "right" but stood out to me, probably because DDL is a little juicy.

packages/web-console/src/scenes/Schema/Table/index.tsx
```diff
@@ -221,9 +239,40 @@ const Table = ({
...
+        const children: TreeNode[] = []
+
+        if (matView && matViewData) {
+          children.push({
+            name: "Base Table",
+            render: () => (
+              <Row
+                table_id={id}
+                kind="info"
+                name="Base Table"
```
goodroot, line 252:
```suggestion
                name="Base table"
```

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+      acc[index].push(table)
+      return acc
+    }, [[], []] as QuestDB.Table[][]).map(tables => 
+      tables.sort((a, b) => a.table_name.toLowerCase().localeCompare(b.table_name.toLowerCase()))
+    )
+
+    return {
+      groups: [
+        { name: 'Tables', count: regularTables.length, expanded: tablesExpanded },
+        { name: 'Materialized Views', count: matViewTables.length, expanded: matViewsExpanded }
```
goodroot, line 108:
```suggestion
        { name: 'Materialized views', count: matViewTables.length, expanded: matViewsExpanded }
```

goodroot: Could also make plurality a condition - <=1 view, > 1 views.

packages/web-console/src/scenes/Schema/Table/index.tsx
```diff
@@ -160,6 +161,24 @@ const Table = ({
...
+    try {
+      const response = await quest.showMatViewDDL(name)
+      if (response && response.type === QuestDB.Type.DQL) {
+        return response
+      }
+    } catch (error: any) {
+      dispatch(
+        actions.query.addNotification({
+          content: (
+            <Text color="red">Cannot show DDL for materialized view '{name}'</Text>
```
goodroot, line 174:
```suggestion
            <Text color="red">Cannot show query for materialized view '{name}'</Text>
```

If agreed, flagging here too.

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+const Loading = () => {
+  const [loaderShown, setLoaderShown] = useState(false)
+  // Show the loader only for delayed fetching process
+  useEffect(() => {
+      const timeout = setTimeout(() => setLoaderShown(true), 1000)
+      return () => clearTimeout(timeout)
+  }, [])
+
+  return loaderShown ? <Loader size="22px" /> : null
+}
```
goodroot, line 65: The loader appears even when fetch is nearly instant.

https://github.com/user-attachments/assets/c68e94f8-7efe-431f-9b83-bb0569f15347

We see that `View.ready` [finishes](https://github.com/questdb/ui/pull/405/files#diff-5b1dd133afb78c093829837e1b2cb95a1d073ea89a622c0f02a55b90bf9ba064R166) once all of these are fetched:

* tables
* wal info
* matviews
 
Even quick ops trigger a load indicator flash due to component mounting.

Maybe... 

1. Track load start time in a ref
2. Compare against current time during load state
3. Only show loader if duration exceeds 1000ms threshold as intended

Also note that tables show loader only for columns not parent:

https://github.com/user-attachments/assets/440d5c8b-1c2b-4382-bf57-079f76323675

goodroot (commented):
Congrats on first PR 😀 - I appreciate the approach & fearlessness.

A few comments, a few visual/functional things, and a clarifying a couple broad areas.

---

The Table & Materialized view sections match base BG colour.

Screenshot 2025-03-13 at 2 39 57 PM

It seems off to me. Can we delineate them in a stylish way?

Maybe so tables & mat view look as "children" within data sources section?

--- 

Symmetry of the toggleable data sources button broke with filter change:

Screenshot 2025-03-13 at 2 45 52 PM

Before, it balanced with table view contents:

Screenshot 2025-03-13 at 2 50 29 PM

--- 

Below are more general thoughts... 🤔 

---

It is not clear how to "use" the materialized view.

We can see them and copy the query. But there isn't any interactivity beyond that.

What is the goal of the current view?

Can we "help" more? 

* Link out to docs somewhere if there's no view? ie. better zero-state?

* Instead of a copy, perhaps create a tab & launch the query on click?

* If "materialized view" tab is active, refresh when appropriate?

Maybe beyond the scope of our first increment, and we might not have answers yet.

That's cool. 😃 But let's figure out where we're going.

---

"Invalid" states are missing.

If a table is dropped, the related mat views do not respond.

Base table row potential space for warning and repair guidance:

Screenshot 2025-03-13 at 2 47 20 PM

There are also other conditions which can invalidate a mat view:

- Schema changes to the base table 
- With deduplication, views must use same grouping keys as the base table's `UPSERT KEYS`

During these states, the view needs a `REFRESH`.

Do we expect to guide through these cases?

---

🫡 Solid push.

packages/web-console/src/scenes/Schema/Table/index.tsx
goodroot: We are pretty conservative re: casing in general.

packages/web-console/src/scenes/Schema/Table/index.tsx
```diff
@@ -221,9 +239,40 @@ const Table = ({
...
 
+          const response = await showMatViewDDL(table_name)
+          if (response && response.data && response.data.length > 0) {
+            children.push({
+              name: "DDL",
+              render: () => (
+                <Row
+                  table_id={id}
+                  kind="info"
+                  name="DDL"
```
goodroot, line 266:
```suggestion
                  name="Query"
```

Perhaps more simple language for less technical personas?

DDL is correct and also jargon.

Up to you - we can use DDL in code, but maybe soften for viewer exposure.

packages/web-console/src/scenes/Schema/Table/index.tsx
```diff
@@ -221,9 +239,40 @@ const Table = ({
...
+                  value={response.data[0].ddl}
+                  copyable={true}
```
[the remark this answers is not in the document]
emrberk: Three issues here:
1 - padding left-right does not match padding top-bottom.
2 - DDL row height does not match the others.
3 - Copy button is small.
4 - No tooltip

I will fix (1) as it is trivial.
However, for (2) and (3):
- If I prioritize making copy button bigger, DDL row height mismatch becomes even more visible.
- If I prioritize making row heights equal, then the copy button becomes smaller.
See these two in action:
Screenshot 2025-03-14 at 11 59 10

Even though the button becomes smaller, I think the first option seems more elegant than the other. Thus, I suggest continuing with this approach. 

About (4), we don't have it in the copy button in Log as well. Also, as we truncate the DDL, I think it's very intuitive that this is the copy button for that value, thus, suggest not adding a tooltip here.
WDYT?

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
emrberk: >columns have uniform shade, look good
>other rows show varying height

These are addressed in https://github.com/questdb/ui/pull/405#discussion_r1995178327.

About
>we have 3x shade levels

Should I consider removing the row shading on hover, for some type of columns? I couldn't get the problem 😅

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
emrberk: `Materialized View (0)` is I think misleading, and `Materialized View (1)` creates redundancy.
I will convert it to lowercase, but I suggest always keeping it as plural as we have the count there. WDYT?

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
emrberk: Delay part is done.
>Also note that tables show loader only for columns not parent:

It is only shown in columns as the API call happens only for the columns content.
Whereas for materialized views, we need to call the API for DDL as well.

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
goodroot: Sounds good! A nit for sure.

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
goodroot: Ah nice! No worries, the latest commits addressed it.

packages/web-console/src/scenes/Schema/Table/index.tsx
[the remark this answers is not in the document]
goodroot: Agreed - current cut is nice and clean. 👍

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
goodroot: Got it - looks very tidy.

packages/web-console/src/scenes/Schema/Table/index.tsx
emrberk: Changed it to "Query" 👍

packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  &:hover {
+    background: ${color("selection")};
+  }
+`
+
+const ArrowIcon = styled(RightArrow)<{ $expanded: boolean }>`
+  color: ${color("gray2")};
+  transform: ${({ $expanded }) => $expanded ? 'rotateZ(90deg)' : 'none'};
+`
```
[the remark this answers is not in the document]
emrberk: Removed the arrow & disabled expanding if there is no table or materialized view 👍

emrberk:
>Can we delineate them in a stylish way?
Maybe so tables & mat view look as "children" within data sources section?

@goodroot please take a look at this one:
Screenshot 2025-03-15 at 23 45 50
The hierarchy seems more visible this time.

emrberk:
Hey @goodroot, I updated the PR with the design we agreed on the discussion.

goodroot (approved):
Solid work - the tree view feels fast and the ux is clean.

In another iteration, it'd be worthwhile to make the mat view section useful.

Right now, it's informational. 

But "open query in tab" or something would go a long way.

---

Appears unrelated, but flagging.

When testing:

* created table & matview towards that table
* deleted table
* created new table
* tried to create new mat view, failed

It put me in a state where I cannot make new matviews: "no matching table."

This is most likely on the DB side. I've raised it there.

goodroot (commented):
Solid work - the tree view feels fast and the ux is clean.

In another iteration, it'd be worthwhile to make the mat view section useful.

Right now, it's informational. 

But "open query in tab" or something would go a long way.

I will have to step away for some time.

So, LGTM from here. 👍

PR #412, opened 2025-04-02
feat(ui): make SSO re-authentication optional on logout

glasstiger:
Currently the web console forces the user to re-authenticate with the OAuth2/OIDC provider after logout by sending the query param `prompt=login` on the authorization code request.

Logout can be initiated in a number of ways:
 - by pressing the Logout button
 - by not having a refresh token when the access token expires
 - by various authentication related issues

In most cases the user has a valid session with the OAuth2/OIDC provider, which means that the provider is able to issue an authorization code without the need for the user typing their password in again.
Therefore the query param `prompt=login` should be optional on logout, and should be sent only when required.
This includes scenarios when:
 - the user logs out to change to another SSO user account
 - authentication attempt was unsuccessful, such as the `state` parameter does not match or missing

However, in scenarios when the user would logout only to switch to a database user, such as built-in admin, do something, and change back to their SSO user, we should not send `prompt=login`.
Now the Login screen will give the option to the user to continue with the currently logged in OAuth2 user, or switch to a different account.


glasstiger:
> Hi @glasstiger,
> 
> I think having `Logout` and `SSO Logout` at the same time is misleading, and not clear at first glance

Yes, UX would have needed more work for sure.

I have merged your changes from https://github.com/questdb/ui/pull/416 into this PR now.

PR #413, opened 2025-04-03
feat(ui): owner setting for tables created during CSV import

glasstiger:
Adding a new dropdown on the CSV import panel for selecting the owner of new tables created during import.
The dropdown lists the user and their groups as options.
Users logged in via SSO will see only their groups, as permissions cannot be granted to external users.

Requires:
https://github.com/questdb/questdb/pull/5464
https://github.com/questdb/questdb-enterprise/pull/579


emrberk (dismissed):
LGTM

emrberk (approved):

PR #417, opened 2025-04-14
fix(web-console): don't use navigator.clipboard in insecure context

emrberk:
Fixes https://github.com/questdb/questdb/issues/5570, we should not try to access the navigator.clipboard in insecure context (e.g. custom domain name in localhost, using http).


javier (approved):
Tested both using IP like 192.168.1.50:9999 and javier:9999 and it works in both cases. It was failing with the web console bundled with 8.2.3

PR #427, opened 2025-04-29
feat(web-console): instance naming UI

emrberk:
User can set the instance name of the database in the web console.

Requires:
https://github.com/questdb/questdb/pull/5621 and https://github.com/questdb/questdb/pull/5689


emrberk:
Telemetry changes LGTM

Issue #346, opened 2024-10-19
Need to improve the suggestion while adding columns in the table. 


AayushSaini101:
We will get a warning when the column name is the same: 

<img width="544" alt="image" src="https://github.com/user-attachments/assets/67e7ba93-5432-466f-ab96-dd165f7e63d8">

Not getting warning when the column are same but in different case,
<img width="579" alt="Screenshot 2024-10-19 at 2 15 30 PM" src="https://github.com/user-attachments/assets/084658b0-7d39-4dff-b108-1d111ef7023f">

The QuestDB creating table is not case sensitive: 
<img width="765" alt="image" src="https://github.com/user-attachments/assets/e7c6c9e1-03e9-4650-a8f6-a5183e8ad17b">

omsuneri:
@AayushSaini101 I'm willing to contribute to this issue please assign it to me

PR #454, opened 2025-07-14
FIXES ISSUE #346 enforce case-insensitive uniqueness for column names

omsuneri:
column names in the table schema dialog are now validated case-insensitively to prevent duplicate columns like 'Age and 'age' this fixes issue #346 where duplicate column names differing only by case were incorrectly allowed

Demo : 

https://github.com/user-attachments/assets/9d213e8a-f19b-4114-80a0-9f3eb5070789

### DIFF
--- a/packages/web-console/src/components/TableSchemaDialog/dialog.tsx
+++ b/packages/web-console/src/components/TableSchemaDialog/dialog.tsx
@@ -157,7 +157,7 @@ export const Dialog = ({
         }
         return value
       })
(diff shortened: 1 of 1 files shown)


omsuneri:
@AayushSaini101 @emrberk the issue is resolved, please review the PR and request if any further changes required !!

emrberk (commented):
A good addition would be trimming the values as well, not to end up with `columnName` and ` columnName ` being used.
Otherwise, this one fixes the case-insensitive columns issue.

PR #461, opened 2025-08-13
fix: outdated version in `local-development-setup`

### LINKED ISSUE #460: `local-development-setup.md` has an outdated version

hamirmahal:
Fixes #460

### DIFF
--- a/docs/local-development-setup.md
+++ b/docs/local-development-setup.md
@@ -15,17 +15,19 @@ Each step is described in below.
 
 There are few ways to clone a repository. Check [Github docs](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository) if you need help.
 
-* Clone this repository using SSH:
+- Clone this repository using SSH:
+
   ```
   git clone git@github.com:questdb/ui.git
   ```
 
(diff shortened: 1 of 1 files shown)


docs/local-development-setup.md
```diff
@@ -39,20 +41,18 @@ It does feel a bit strange to have dependencies committed, but it's a well worth
...
 ### 2. Bootstrap dependencies
 
-
 #### `node` and `yarn`
 
 First make sure you have `node` and `yarn` and that their versions are compatible with this project.
 This is a common source of issues, best not to skip this step:
 
-* `node -v` should return `16.13.1`\
+- `node -v` should return a version compatible with what's in [.nvmrc](../.nvmrc)\
```
hamirmahal, line 49: Rather than hard-coding the version here, and worrying about it getting out of date, we can just point users directly to `.nvmrc`, which should have the version of `node` used by the project.

hamirmahal, line 49: Rather than hard-coding the version here, and worrying about it getting out of date, we can just point users directly to `.nvmrc`, which should have the version of `node` used by the project.

PR #474, opened 2025-09-23
fix(web-console): handle timestamp_ns format in CSV imports

emrberk:


javier:
Ingestion works with nanos now, but I cannot override `timestamp_ns` to `timestamp`. I ingest via csv with the defaults, all good (no designated ts, no proper datatypes, but it inserts everything, which failed before). Now I want to override the schema, so I open the file again and click schema. I changed the data types, but I am feeling playful and even if I have nanos on the CSV, I want to cast to timestamp in micros, so I choose `timestamp` for the data type. The data is ingested, but the type is actually `timestamp_n`s, which was not what I asked. I can see this is passing the type to the backend as `TIMESTAMP_NS`.

Even if passed properly, the backend is also ignoring it (but if the table already exists, it ingest with the proper type), but the console should pass the desired datatype, rather than timestamp_ts

PR #476, opened 2025-10-01
feat(web-console): Add column name search to Schema filter

### LINKED ISSUE #363: table filter could search in columns too
The filter in the web console's left bar could search not just in table names but also in columns.

This would be quite nice when you have tables with a huge number of columns. 

Screenshot From 2024-12-17 15-21-24

Itz-Agasta:
Enhanced the table filter in the Schema panel to search both `table names` & `column names`. When a column matches the search query, the table automatically expands (in-memory only) to show the matching columns.

### Changes
- **Filter logic**: Updated table filtering to include column name matches using existing Redux `allColumns` data
- **Auto-expansion**: Tables with matching columns now auto-expand their table row AND columns folder during search
- **State management**: Search-driven expansions are purely in-memory and never persist to localStorage
- **Race condition fix**: Added cleanup mechanism (`isCancelled` flag) to prevent stale async operations from causing incorrect expansion states during rapid input changes

### Implementation Details
1. **Column matching**: Uses `allColumns` from Redux store (already fetched via `information_schema.columns()`) - no additional network requests
2. **`tablesWithColumnMatches` Set**: Computed via `useMemo` to track which tables need auto-expansion
3. **In-memory expansion**: Sets `node.isExpanded = true` directly on tree nodes without calling `setSectionExpanded()` (localStorage)
4. **Async safety**: Effect cleanup function prevents race conditions when user types/deletes rapidly

### Behavior
- [X] Search for table name → table stays collapsed (existing behavior preserved)
- [X]  Search for column name → table + columns auto-expand, matching columns visible
- [X]  Clear search → all expansions revert to localStorage state (user's manual expansions preserved)
- [X]  Rapid typing/backspace → no ghost expansions

### Testing
Tested with sample tables containing columns like `col_string`, `col_ts`, `sensor_id`:
- Column search correctly expands matching tables
- Table name search maintains collapsed state
- Rapid input (backspace spam) correctly handles state
- Clearing search restores previous expansion states

https://github.com/user-attachments/assets/21e8b5b6-c64c-4af7-911b-5cda3db4f855

Closes #363

### DIFF
--- a/packages/browser-tests/cypress/integration/console/schema.spec.js
+++ b/packages/browser-tests/cypress/integration/console/schema.spec.js
@@ -1,6 +1,6 @@
 /// <reference types="cypress" />
 
-const contextPath = process.env.QDB_HTTP_CONTEXT_WEB_CONSOLE || ""
+const contextPath = process.env.QDB_HTTP_CONTEXT_WEB_CONSOLE || "";
 const baseUrl = `http://localhost:9999${contextPath}`;
 
 const tables = [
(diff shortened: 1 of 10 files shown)


Itz-Agasta:
Hi @emrberk  Hope y are having a great day, 
So, about memoization for column indexing  i was talking about in the issue... so, my idea was to pre-compute a searchable index of `table→columns` mapping when tables/columns change, rather than accessing the Redux store for each table during every search keystroke.

For schemas with many tables, this could avoid repeated Redux store lookups. But while working on the issue, going through  the repo, I found that ithis would be unnecessary tbh. As the `allColumns` data comes from `selectors.query.getColumns` which is already memoized by Redux's selector pattern. The columns are fetched once via `information_schema.columns()` and stored in the Redux store.

Also my current implementation uses `useMemo` for both the filtered tables list and the `tablesWithColumnMatches` Set, so the column matching computation only runs when `tables`, `query`, or `allColumns` change... exactly when it needs to.

Issue #331, opened 2024-09-17
Default ready to use SELECT TOP query


enricobenedos:
### Is your feature request related to a problem?

We think that one of the common operations users/developers want to execute on the database during developing phases, but also on production, is a fast view of table records. At the moment we need to rewrite always the same base query that is always:
`SELECT * FROM my_beautiful_table`. And then add something like "ORDER BY timestamp DESC".

This is a little bit annoying compared to other databases GUI clients.

### Describe the solution you'd like.

We would like to have a new item in every table context menu like "Select TOP 1000 rows" (like Microsoft MSSQL) or "View Data" (like DBeaver). That can simply generate the query and add it to the text editor.

The additional feature could be having two more options when a table have a timestamp index that permits to select first top n items in ascending or descending order.

### Full Name:

Enrico Benedos

### Affiliation:

Deimos

enricobenedos:
Thank you @nwoolmer 

Mine is only a suggestion because during the week, me and my colleagues, rewrite always the same stuff that we don't need to do with other clients . The most common thing to do before write the real query is always the same stupid select to take a look at the data 🥲

TSP06:
hey @enricobenedos i would like to work on this

enricobenedos:
Hello @TSP06, your PR are welcome 👍🏻

HVbajoria:
Hey @enricobenedos ,

I would like to work on this!

HVbajoria:
Hi @enricobenedos,

Sorry for the issue with the last PR.

Trying to learn from my mistakes. 

Could you please check once and assist me if it is correct this time so that I can update things if required: https://github.com/questdb/ui/pull/480

Thanks a lot!

enricobenedos:
Hi @HVbajoria,
it is nice for me. It is exactly what I mean.

Hope @nwoolmer team / questdb team can validate and approve this PR.

@PriyanshuRj think that it has been solved by @HVbajoria PR.

PR #477, opened 2025-10-01
Schema: enhance table filter to search table and column names with safe in-memory auto-expansion

HVbajoria:
### Description
Enhanced the table filter in the Schema panel to search both table names and column names. When a column matches the search query, the table automatically expands (in-memory only) to show the matching columns.

### Changes

- **Filter logic:** Updated table filtering to include column name matches using existing Redux allColumns data.
- **Auto-expansion:** Tables with matching columns now auto-expand their table row AND columns folder during search.
- **State management:** Search-driven expansions are purely in-memory and never persist to localStorage.
- **Race condition fix:** Added cleanup mechanism (isCancelled flag) to prevent stale async operations from causing incorrect expansion states during rapid input changes.

### Implementation Details

- **Column matching:** Uses `allColumns` from the Redux store (already fetched via `information_schema.columns()`); no additional network requests.
- **`tablesWithColumnMatches` Set:** Computed via `useMemo` to track which tables need auto-expansion.
- **In-memory expansion:** Sets `node.isExpanded = true` directly on tree nodes without calling `setSectionExpanded()` (which writes to `localStorage`).
- **Async safety:** Effect cleanup function prevents race conditions when users type/delete rapidly.

### Behavior

- **Search for table name** → table stays collapsed (existing behavior preserved).
- **Search for column name** → table + columns auto-expand, matching columns are visible.
- **Clear search** → all expansions revert to localStorage state (user’s manual expansions preserved).
- **Rapid typing/backspace** → no ghost expansions.

### Testing

Tested with sample tables containing columns like col_string, col_ts, sensor_id:
- Column search correctly expands matching tables.
- Table name search maintains collapsed state.
- Rapid input (backspace spam) correctly handles state without stale expansions.
- Clearing search restores previous expansion states.

### Video
https://github.com/user-attachments/assets/91008eb0-252a-49ae-ac5d-f049838d93fb

### DIFF
--- a/packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
+++ b/packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
@@ -12,7 +12,7 @@ import { createTableNode, createColumnNodes, getNodeFromSchemaTree, updateAndGet
 import { useRetainLastFocus } from "./useRetainLastFocus"
 import { getSectionExpanded, setSectionExpanded, TABLES_GROUP_KEY, MATVIEWS_GROUP_KEY } from "../localStorageUtils";
 import { useSchema } from "../SchemaContext";
-import { QuestContext } from "../../../providers";
+import { QuestContext, useEditor } from "../../../providers";
 import { PartitionBy } from "../../../utils/questdb/types";
 import { useDispatch } from 'react-redux';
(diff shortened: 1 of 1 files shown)


HVbajoria:
Resolves the issue: #331

PR #480, opened 2025-10-03
Add "View Data" context menu options for quick table querying

HVbajoria:
Implements quick data viewing functionality requested in the issue #331 to reduce repetitive query writing.

**Changes:**
- Added "View data (Top 1000)" context menu item for all tables
- Added timestamp-based viewing options ("View latest/earliest by timestamp") for tables with designated timestamps
- Automatically generates and appends `SELECT * FROM table_name LIMIT 1000` queries to the editor
- Includes `ORDER BY timestamp DESC/ASC` for timestamp-based queries

**Addresses:**
- Eliminates need to manually write `SELECT * FROM my_beautiful_table` queries
- Provides quick access to recent/oldest records via timestamp ordering

Users can now right-click any table in the schema tree and immediately view data without manual query construction.

Reference Video:

https://github.com/user-attachments/assets/ad50c878-4900-4122-ae39-99a01d1d33ea

### DIFF
--- a/packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
+++ b/packages/web-console/src/scenes/Schema/VirtualTables/index.tsx
@@ -12,7 +12,7 @@ import { createTableNode, createColumnNodes, getNodeFromSchemaTree, updateAndGet
 import { useRetainLastFocus } from "./useRetainLastFocus"
 import { getSectionExpanded, setSectionExpanded, TABLES_GROUP_KEY, MATVIEWS_GROUP_KEY } from "../localStorageUtils";
 import { useSchema } from "../SchemaContext";
-import { QuestContext } from "../../../providers";
+import { QuestContext, useEditor } from "../../../providers";
 import { PartitionBy } from "../../../utils/questdb/types";
 import { useDispatch } from 'react-redux';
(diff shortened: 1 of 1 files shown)


HVbajoria:
Hi @emrberk ,

I have tried implementing the suggestion.

Could you please check once?

https://github.com/user-attachments/assets/f7c4180a-d32b-47ad-9b19-51fa62a3e06c

emrberk:
@HVbajoria
- Build is broken in your branch.
- As per the attached video, the hover ergonomics is not working as expected. When cursor is placed on "Queries", the focus should not be lost.
- The main context menu should not lose its z-axis positioning when Queries is open.
- The feature needs new tests to be added in `schema.spec.js`. 

Please make sure everything functions correctly and the PR is ready before requesting a review.

HVbajoria:
> @HVbajoria
> 
> * Build is broken in your branch.
> * As per the attached video, the hover ergonomics is not working as expected. When cursor is placed on "Queries", the focus should not be lost.
> * The main context menu should not lose its z-axis positioning when Queries is open.
> * The feature needs new tests to be added in `schema.spec.js`.
> 
> Please make sure everything functions correctly and the PR is ready before requesting a review.

Hi @emrberk,

Could you please help me with:
-  How to create the build?
- Could you please explain me the 2nd and 3rd points?

PR #544, opened 2026-03-18
feat: add web console usage telemetry

emrberk:
Adds anonymous usage telemetry to the web console. Events are stored locally in IndexedDB and flushed to the server in batches.

Telemetry infrastructure (ConsoleEventTracker module)

- `trackEvent(name, payload?)`: fire-and-forget entry point; writes to IndexedDB via Dexie (db.ts)
- `sendPipeline.ts`: background flush loop that runs every 5 s when there are pending events
- Calls `console-events-config` to get the server-side cursor (lastUpdated), then POSTs only newer events via `add-console-events`
- Exponential back-off (base 1 s, up to 2^9 s) with independent retry counters for the check and send phases
  - `MAX_BATCH_SIZE` = 1 000 events per POST; successfully sent entries are deleted from IndexedDB immediately
  - On startup, `trimToMaxRows(10 000)` evicts the oldest events to bound local storage


jerrinot (dismissed):
this looks good, but it should not be merged before https://github.com/questdb/serverless-functions/pull/68

jerrinot (approved):

PR #546, opened 2026-03-26
refactor: allow table details empty state with table selection

emrberk:
- Adds a table selector to table details drawer header if there is no table details history
- Uses the same selector as metrics table selector
- Styling updates for metrics header (table name, title, actions)
- Layout updates to metrics dashboard

Screenshot 2026-03-26 at 10 17 39


javier (changes requested):
Very cool. Works well, but I just discovered an issue that was already broken, just we didn't notice. Regular views don't appear on the details tab and cannot be selected, but they still have the "show details" option on hover on the left pane. I think the ideal would be that regular views would have details, but not monitoring tab, so the table creation details could still be seen. But not super strong feelings if we decide not to show details, but right now it is broken

javier (dismissed):

javier (approved):

PR #548, opened 2026-04-10
feat: add license compliance checks to CI

bluestreak01:
## Summary
- Adds a **license-check** CI job that runs in parallel with the existing CI flow and fails if any production npm dependency has a license incompatible with Apache 2.0 (allowlist: MIT, ISC, Apache-2.0, BSD-2-Clause, BSD-3-Clause, 0BSD, Unlicense, CC0-1.0, CC-BY-3.0, CC-BY-4.0, Python-2.0, BlueOak-1.0.0, Zlib)
- Adds a **font license check** that requires every font directory (`src/styles/fonts/`, `public/fonts/`) to contain a LICENSE file — currently passes for Open Sans (Apache 2.0) and **intentionally fails for PP Formula** which has no license file
- Production builds now generate `dist/THIRD_PARTY_LICENSES.txt` containing full license texts for all bundled npm dependencies and fonts
- Adds `yarn check:licenses` script for local use

### New dependencies
- `license-checker` — scans npm dependency licenses against an allowlist
- `rollup-plugin-license` — extracts license texts into `THIRD_PARTY_LICENSES.txt` at build time

### Known issue
The `public/fonts/PP Formula` font has no LICENSE file. The CI check will fail until this is resolved — either by adding a valid license file, or by replacing/removing the font. This is intentional to surface the compliance gap.

## Test plan
- [x] `yarn check:licenses` passes for npm deps, fails for missing PP Formula license
- [x] `yarn build` produces `dist/THIRD_PARTY_LICENSES.txt` with 206 package entries + Open Sans font license
- [ ] CI `license-check` job runs in parallel with existing CI job


joshtriplett:
> currently passes for Open Sans (Apache 2.0)

Open Sans is not Apache-2.0, it's OFL 1.1, which is Open Source, but not a subset of Apache-2.0 (it restricts the use of the font name, and it requires distributing under the same license rather than under any other license). It's a perfectly reasonable Open Source license, it just needs the license file included.

emrberk:
Hey @joshtriplett,

Now Open Sans license is correctly added as OFL 1.1, PPFormula font is removed, a build plugin to check/copy licenses is added. License violations will result in build failures.

Thanks for reporting the issue. Let us know if there are any remaining concerns.

joshtriplett:
Thank you! This looks great. And I really appreciate you adding checks to CI to keep this robust. :heart:

emrberk (approved):

PR #556, opened 2026-04-21
fix: remove implicit select suggestions filtering, context-aware function suggestions

emrberk:
Related parser commit: https://github.com/questdb/sql-parser/commit/5a422b098fef678238bcb12fb1e71a59f42ba7e3

## Summary                                                                                                                                                                     
                                                                                                                                                                               
Bumps `@questdb/sql-parser` from 0.1.8 to 0.1.11 — autocomplete now picks function categories from the cursor's grammar context instead of dumping the full function list at every identifier position, implicit select statements suggestions are included in the statement start.
Fixes hanging editor when there is a trailing comma in the select list.


Adds manual resolutions for `lodash` and `lodash-es` to version `4.18.1` to resolve vulnerability scan issues
                                                                                                                                                                                 
  ## Examples     

  **`t|`**
  - Before: Not suggesting any tables
  - After: Suggests `trades` (if table exists), `tables`, `wal_tables` etc.                                                                                                                                                
   
  **`SELECT * FROM trades ASOF JOIN m|`**                                                                                                                                        
  - Before: 332 entries — `materialized_views`, `maxUncommittedRows`, `max`, `median`, `mid`, `millis`, `min`, …                                                               
  - After: 3 entries — `materialized_views`, `memory_metrics`, `table_writer_metrics`                                                                                            
                                                                                                                                                                                 
  **`SELECT * FROM trades WHERE price = c|`**                                                                                                                                    
  - Before: scalars + aggregates mixed (`coalesce`, `count`, `count_distinct`, `corr`, `covar_*`, …)                                                                             
  - After: scalars only (`coalesce`, `concat`, `cos`, `ceil`)                                                                                                                    
                                                                                                                                                                                 
  **`INSERT INTO trades VALUES (n|`**                                                                                                                                            
  - Before: 0 suggestions                                                                                                                                                        
  - After: `now`, `now_ns`, `nullif`, `nanos`, `netmask`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
 
       - uses: actions/setup-java@v4
         with:
-          java-version: "17"
+          java-version: "25"
           distribution: "temurin"
           cache: maven
 
--- a/.github/workflows/tests_with_context_path.yml
+++ b/.github/workflows/tests_with_context_path.yml
@@ -18,7 +18,7 @@ jobs:
 
       - uses: actions/setup-java@v4
         with:
-          java-version: "17"
+          java-version: "25"
           distribution: "temurin"
           cache: maven
 
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "@monaco-editor/react": "^4.7.0",
     "@phosphor-icons/react": "^2.1.10",
     "@popperjs/core": "2.4.2",
-    "@questdb/sql-parser": "0.1.8",
+    "@questdb/sql-parser": "0.1.9",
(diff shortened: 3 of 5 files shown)


javier (commented):
I still get some weird results from postgresql meta functions

Image

Also, not sure if intended behaviour, but when I search from some letter and I have partial results, I get partial results only if the letter is at the beginning of a word (beginning of word, after period, after underscore), but not if the letter is midword

javier (commented):
I see the opposite behaviour to what described for aggregates. This is on older version from main

Image

and this is in newer version from this PR

Image

javier (approved):

PR #582, opened 2026-07-01
fix(console): auto-collapse sidebars on narrow viewports

midaksh:
Part **B** of #574: when the viewport shrinks, auto-collapse the left and right sidebars so the notebook stays the primary, usable surface.

Stacks on #581 (Part A).

### Changes
**`src/hooks/useScreenSize.tsx`**
- Add `md` (1024px) and `lg` (1280px) breakpoints alongside the existing `sm` (767px)
- Same throttled resize + `shallowEqual` update pattern as `sm`

**`src/scenes/Console/index.tsx`**
- **Right panel** (`AI` / table details): hide below **1280px** (`!lg`) — larger footprint (470px min), collapses first
- **Left panel** (schema / search): hide below **1024px** (`!md`) — in addition to existing `!sm` at 767px
- Set `proportionalLayout={false}` on the outer `Allotment` so the AI panel restores its saved `preferredSize` after resize instead of expanding to max width

### Behaviour

| Viewport | Right panel | Left panel | Notebook |
|---|---|---|---|
| > 1280px | visible if open | visible if open | normal layout |
| 1025–1280px | auto-hidden | visible if open | gains ~470px |
| 768–1024px | auto-hidden | auto-hidden | fills center |
| ≤ 767px | auto-hidden | auto-hidden | same as before (`sm`) |

Panel open/close state (Redux `activeSidebar`, localStorage `leftPanelState`) is **not** mutated — panels restore when the viewport widens past the threshold.

### Out of scope
- Mobile icon visibility (datasource hidden at `sm`) – pre-existing
- Drawer/overlay for panel toggles on very narrow viewports – **follow-up if desired**

- Closes: #574 (Part B)
- Stacked on: #581

### DIFF
--- a/src/hooks/useScreenSize.tsx
+++ b/src/hooks/useScreenSize.tsx
@@ -38,19 +38,32 @@ type Props = Readonly<{
 
 type ScreenSize = Readonly<{
   sm: boolean
+  // True below 1024 px: left schema/search panel auto-collapses so the
+  // notebook keeps a usable width even when the right panel is also open.
+  md: boolean
+  // True below 1280 px: right AI/table-details panel (470 px min-width)
+  // auto-collapses first because it has the larger footprint. The left panel
+  // remains available at 1024–1279 px.
+  lg: boolean
 }>
 
 enum Breakpoint {
   SM = 767,
+  MD = 1024,
+  LG = 1280,
 }
 
 const ScreenSizeContext = createContext<ScreenSize>({
   sm: window.innerWidth <= Breakpoint.SM,
+  md: window.innerWidth <= Breakpoint.MD,
+  lg: window.innerWidth <= Breakpoint.LG,
 })
 
 export const ScreenSizeProvider = ({ children }: Props) => {
   const [screenSize, setScreenSize] = useState<ScreenSize>({
     sm: window.innerWidth <= Breakpoint.SM,
+    md: window.innerWidth <= Breakpoint.MD,
+    lg: window.innerWidth <= Breakpoint.LG,
   })
   const [_screenSize, _setScreenSize] = useState<ScreenSize | undefined>()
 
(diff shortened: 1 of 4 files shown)


emrberk:
Hey @midaksh,

I think this is a great start with the breakpoints!
My concern is that the users explicitly wants to open left/right sidebars in md and sm screens will not be able to do that anymore. How about the following approach:
- When we shrink from `lg` to `md`, if both sidebars are open, close right sidebar automatically, and only show the left sidebar. The user will be able to reopen it any time they want. We don't block it.
- When we shrink from `md` to `sm`, close all sidebars automatically. The user will be able to reopen any sidebar if they want.
- Let's not hide "Data sources" toggle in left sidebar in `sm` view. This was a preexisting issue.

Let me know if you need any help!

midaksh:
> Hey @midaksh,
> 
> I think this is a great start with the breakpoints! My concern is that the users explicitly wants to open left/right sidebars in md and sm screens will not be able to do that anymore. How about the following approach:
> 
> * When we shrink from `lg` to `md`, if both sidebars are open, close right sidebar automatically, and only show the left sidebar. The user will be able to reopen it any time they want. We don't block it.
> * When we shrink from `md` to `sm`, close all sidebars automatically. The user will be able to reopen any sidebar if they want.
> * Let's not hide "Data sources" toggle in left sidebar in `sm` view. This was a preexisting issue.
> 
> Let me know if you need any help!

@emrberk This was my actual idea to start with this approach, but since it was a pre-existing logic for `sm`, I hesitated to make the logical changes, but now as you have suggested I will implement it shortly and notify you.

midaksh:
Hello @emrberk all the instructions that were provided have been incorporated, let me know if something else is needed.

emrberk:
Hey @midaksh,

All looks good, thanks for your contribution!

PR #588, opened 2026-07-16
fix: preserve query result text

### LINKED ISSUE #587: Web Console decodes HTML entities in results, corrupting displayed and copied values
**Describe the bug**

The web console runs every result cell through `unescapeHtml()` before rendering, and the copy paths reuse the same text. But the server does not HTML-encode values in `/exec` JSON responses, so any stored value that legitimately contains an entity sequence (`&amp;`, `&lt;`, `&gt;`, `&quot;`, `&#039;`, `&nbsp;`) is silently altered, both in the grid display and in the copied value.

This affects the legacy grid (`src/js/console/grid.js`), the new ResultGrid (`src/components/ResultGrid/inlineGridUtils.ts`), and the markdown copy (`src/components/ResultGrid/resultPageMarkdown.ts`).

**Steps to reproduce**

On https://demo.questdb.io (QuestDB 9.4.3, console 2.0.0):

```sql
SELECT '&amp;' AS literal_entity, 'a=1&amp;b=2' AS url_query, '&lt;tag&gt;' AS escaped_html;
```

The grid shows `&`, `a=1&b=2` and `<tag>` instead of the stored values. Copying a cell copies the altered value too.

Image

The server returns the values unmodified, so the decode has nothing to undo:

```
$ curl 'https://demo.questdb.io/exec?query=SELECT%20%27%26amp%3B%27%20lit'
{"query":"SELECT '&amp;' lit", ..., "dataset":[["&amp;"]],"count":1}
```

(Same raw output with `version=1` and `version=2` request params.)

**Where it came from**

- #519 correctly switched cell rendering from `innerHTML` to `textContent`. As a side effect that also fixed #320, since the cell copy path now reads `textContent`.
- #525 then added `unescapeHtml()` on top, on the assumption that the server sends HTML entities. As far as I can tell from `/exec` responses on 9.4.3 it doesn't, so the unescape only ever fires on user data that happens to look like an entity.

Realistic hits: URL query strings stored after HTML-escaping upstream, HTML fragments in varchar columns, scraped or event payloads.

If some older server versions do escape entities in query responses (which would explain #525), the unescape would need to be gated on that rather than applied unconditionally.

Happy to submit a PR removing the unconditional unescape (grid.js, inlineGridUtils, resultPageMarkdown, plus their tests) if you can confirm the intended behavior.

emrberk:
## Summary

- remove HTML entity decoding from the modern and legacy result grids
- preserve result values exactly in rendered cells, clipboard output, and Markdown export
- preserve leading whitespace so plain-text EXPLAIN plans keep their hierarchy
- update result-grid and Markdown tests for literal HTML-entity-looking values

## Problem

QuestDB's HTTP EXPLAIN response historically contained HTML entities as a Web Console presentation workaround. The console compensated by applying `unescapeHtml` to every result value, not only to structural query-plan indentation.

That made ordinary query results lossy: legitimate values such as `&nbsp;`, `&lt;`, `&gt;`, `&amp;`, `&quot;`, and `&#039;` could be changed even though they were user data. Trying to detect EXPLAIN queries client-side would still be fragile and would not resolve collisions between plan structure and SQL literals embedded in a plan.

Closes #587.

## Solution

The paired QuestDB server change returns EXPLAIN plans as plain text. With that contract, the Web Console can treat every response value as data:

- remove all `unescapeHtml` calls and the unused helper
- preserve whitespace in both grid implementations with `white-space: pre`
- leave query-plan Markdown rows untouched
- keep entity-looking strings literal in display, copy, and export paths

No client-side EXPLAIN detection or SQL tokenization is needed.

## Validation

- focused ResultGrid unit tests: 40 passed
- TypeScript typecheck passed
- relevant ESLint checks passed
- production build passed
- manually validated ordinary entity-like values, real leading spaces, raw angle brackets, EXPLAIN indentation, comparison operators, SQL literals inside EXPLAIN, and Markdown copy against the paired development server

## Related

- QuestDB server fix: https://github.com/questdb/questdb/pull/7406
- Original issue: https://github.com/questdb/ui/issues/587


emrberk:
# Review — PR #588 "fix: preserve query result text" (Level 3)

**Quality gate:** typecheck ✅ · lint ✅ · build ✅ · unit tests ✅ (1510 passed, incl. the new `toSingleLineDisplay` and updated markdown/copy tests).

## Issues

| # | Issue | Category | Severity | Location | Description | Repro | Suggested fix |
|---|-------|----------|----------|----------|-------------|-------|---------------|
| 1 | EXPLAIN plans show literal `&nbsp;`/`&lt;` on un-upgraded servers | Query execution & data integrity | Moderate | in-diff (display path now emits server text verbatim; depends on paired server contract) | The fix's correctness relies entirely on the paired server change ([questdb#7406](https://github.com/questdb/questdb/pull/7406)) returning EXPLAIN plans as plain text. Against a server that hasn't shipped it, EXPLAIN indentation (`&nbsp;`) and operators (`&lt;`,`&gt;`) now render as literal entity text in the grid, and copy/Markdown export the same broken text. Confirmed `demo.questdb.io` — the `/exec` proxy target in the modified `vite.config.mts` — **still entity-encodes**: `dataset:[["&nbsp;&nbsp;filter: 100&lt;amount"], ...]`. There is no `buildVersion` guard gating the behavior. In the normal bundled binary the console ships with its own server so both halves land together; the risk is decoupled deployments (this dev branch → demo, QuestDB Cloud console served independently of the engine, or a standalone console pointed at an older instance). | • Run `EXPLAIN SELECT * FROM trades WHERE amount > 100` in the Result grid or a notebook cell against any server without questdb#7406 (e.g. this branch's own dev server → demo.questdb.io) • Plan renders as `&nbsp;&nbsp;filter: 100&lt;amount` instead of indented `  filter: 100<amount` | Confirm questdb#7406 is released everywhere this console build can reach before merging, and coordinate the release. (Client-side re-scoping the unescape to the `QUERY PLAN` column is *not* a clean fix — it would re-corrupt entity-looking SQL literals embedded in a plan, which is exactly why the server-side fix was chosen.) |
| 2 | Embedded tabs now render as wide tab-stops | Styling & theming | Minor | in-diff (`styles.ts` `CellText`, `_grid.scss` `.qg-r`: `nowrap`→`pre`) | `toSingleLineDisplay` collapses only `[\r\n]`, not `\t`. Under the old `white-space: nowrap` a tab collapsed to normal inter-word spacing; under `white-space: pre` a VARCHAR/STRING value containing a literal tab now advances to the next tab stop, showing a visibly wide gap in both grids. Cosmetic only — no data loss (copy preserves the raw tab), fixed row height clips overflow. Arguably intended, since the PR's goal is to preserve whitespace. | • Query `SELECT 'a' \|\| chr(9) \|\| 'b'` • Cell shows a wide tab gap instead of a single space | If undesired, collapse tabs too: `text.replace(/[\r\n\t]+/g, " ")` in `toSingleLineDisplay`, plus a test case. Otherwise document as intended. |
| 3 | Legacy copy fallback drops newlines | Async/structure | Minor | in-diff (`src/js/console/grid.js:1716`) | In `copyActiveCellToClipboard`, `valueToCopy` is seeded from `focusedCell.textContent`, which #588 made a newline-collapsed string (`toSingleLineDisplay`). The primary branch correctly re-derives the raw value via `getDisplayedCellValue`, but if `data[pageIndex][rowInPage]` is absent the code falls back to the collapsed `textContent` — contradicting the adjacent comment ("must not leak into the clipboard"). Before #588 the fallback still carried raw newlines. Practically unreachable: a focused/visible cell's data page is loaded, so the fallback shouldn't fire under real interaction. | • Focus a cell whose data page has been evicted, then Cmd+C (not reproducible through normal scrolling/focus) | Seed the fallback from the raw datum, or accept the current behavior given it's unreachable — low priority. |

## False-positives

- **React correctness / async / hooks** — No hooks, deps arrays, refs, timers, subscriptions, or memo comparisons changed. `GridCell` (`React.memo`) already computed `displayValue` inline in render before the PR; swapping `unescapeHtml` for `toSingleLineDisplay` at the same site changes nothing about render behavior. Clean.
- **U+2028/U+2029 (and NEL) render as line breaks under `white-space: pre`** — Empirically false in Chromium: cell height is identical to a plain single line (17px) for these chars; only `\n`/`\r` break (and those are stripped by `toSingleLineDisplay`). Not a regression.
- **XSS reintroduced by removing `unescapeHtml`** — Both grids write the value as a text node (`cell.textContent` / JSX `{displayValue}`), never `innerHTML`. The deleted helper was a *display* transform (added in #525), never a security control. The only `escapeHtml` use left is `quick-vis.ts` SlimSelect labels, untouched and still correct.
- **Copy/Markdown output re-rendered as HTML → entity or markup corruption** — Every `buildResultPageMarkdown`/`formatCellValueForCopy` consumer (main Result panel, notebook actions bar, legacy grid, Cmd+C) pipes straight to `copyToClipboard`. No path feeds grid copy into `react-markdown`/the AI assistant. Clipboard is the terminal sink. All callsites SAFE.
- **Width-sampling regression** — `sampleColumnWidths` uses `formatCellValue` length; newline→space is length-neutral, and display no longer shortens `&amp;`→`&`, so sampled width now matches display *better* than before. Neutral/improvement, not a bug.
- **`toSingleLineDisplay` needs a non-string guard** — `getDisplayedCellValue` provably returns a string for all non-null cell data (null is handled earlier), and every TS caller passes a string. Adding `String(text)` would be defensive dead-weight; dropped per the repo's minimalism.

## Summary

**Verdict: Approve, pending one confirmation.** The change is correct, well-scoped, and well-tested. Removing `unescapeHtml` is a genuine fix — verified that server data arrives raw (a literal `'a&amp;b'` returns `a&amp;b`, which the old console corrupted to `a&b`), and both grids render/copy via text nodes so there's no XSS surface. Display-vs-copy semantics (collapse newlines for display, preserve them for copy/export) are consistent across the React grid, the legacy grid, and the Markdown paths.

**The one thing to resolve:** Finding #1 — the console now depends on the paired server change (questdb#7406) for EXPLAIN rendering, and the demo/dev server this branch proxies to hasn't shipped it, so EXPLAIN plans visibly regress there. This is a documented, intentional tradeoff, but merge should be coordinated with the server release (and reviewers should be aware the local dev experience against demo will show entity-encoded plans until then).

**Regressions/tradeoffs:** old-server EXPLAIN rendering (Finding #1); tabs now render wider (Finding #2, likely intended).

**Verification tally:** 3 findings verified (1 Moderate, 2 Minor); 6 draft findings dropped as false positives (verified against source + empirical browser/curl tests). **In-diff vs out-of-diff: 3 in-diff, 0 out-of-diff** — Agent 12 walked all six consumer callsites (ResultGridAdapter, ResultActionsBar, ResultGridPanel/InlineResultTable, useGridKeyboardNav, legacy grid.js, quick-vis) and found every one SAFE; the changed symbols are pure display/copy utilities whose output only reaches the DOM as text or the clipboard, so a zero out-of-diff count is correct here, not an underrun.

bluestreak01 (approved):

PR #595, opened 2026-08-03
feat: notebook-level auto-refresh for charts and grids with per-cell overrides

emrberk:
**Tandem PR**: https://github.com/questdb/mcp-bridge/pull/2

Brings **auto-refresh to grid (run-mode) cells** — previously a chart-only capability — and adds a **notebook-level auto-refresh default** that every cell showing a chart or a grid inherits and can override per cell. Nothing polls unless the cell or the notebook says so — the terminal fallback is Off for every view.

## What's changed

- **Auto-refresh for grid cells (new).** Before this PR, run-mode cells never polled. Grids now refresh per statement: each slot swaps its own rows in place, a failing statement keeps its old rows and carries its error, and siblings refresh independently. Cells containing DDL/DML never poll — auto-refresh never executes a write. Grids stay Off until you opt in.
- **Notebook default** — a toolbar control sets one cadence (Auto / Off / `1s` / `5s` / `10s` / `30s` / `1m`) for the whole notebook. Auto adapts the interval to response time (2s–60s).
- **Per-cell overrides** — each cell's refresh dropdown gains a "Notebook default" (inherit) option, on charts and grids alike. An override is marked with a dot and always wins over the default. "Reset cell overrides" clears them all in one action.
- **Refresh cells** — one click refreshes every chart and grid, superseding in-flight background polls.
- **Behavior change: drawing a chart no longer implies auto-refresh.** `autoRefresh` has one writer (the user) and one meaning — a stored value is a user choice. A new chart inherits the notebook default, so with no default set it renders once and stands still until you opt in.

## Migration

One-time, on load: legacy draw cells that polled through the old implicit per-view Auto get an explicit, resettable `Auto` override, so existing dashboards keep refreshing after the upgrade. No notebook default is ever synthesized, and grids stay off until you opt in.

## AI / MCP

- New `set_notebook_autorefresh` tool sets the default; `reset_cell_overrides` clears per-cell values in the same atomic call.
- `set_cell_autorefresh` accepts `null` to clear an override; it applies to charts and grids alike.
- `apply_notebook_state` gains `auto_refresh_default`; snapshots report the effective default, `auto_refresh_blocked` for write cells, and `last_refresh_error` per cell.

## Reliability

- The configured interval holds across hide/reveal and tab refocus — no query burst on every alt-tab.
- Grid frames persist with their refresh errors, throttled to one write per 10s window; the throttle is bypassed only when the error set changes or on manual refresh, so a persistently failing statement no longer writes to IndexedDB every tick while a reload always shows the current state and a fresh `savedAt`.
- With the editor hidden (view-maximized), Cmd+Enter runs the statement of the selected tab — including a "Not run" tab — never a stale result index.


emrberk:
# PR #595 Review — Level 3

Scope: branch `feat/notebook-level-refresh-interval` at commit `51e1fc78`.
Diff against `main`: 79 files, +9685 / −2947.
Depth: level 3 — full 13-agent pass, every finding verified against source.

## Quality checks

| Check | Result |
|---|---|
| `yarn typecheck` | ✅ PASS |
| `yarn lint` | ✅ PASS |
| `yarn test:unit` | ✅ PASS (1978 tests, 87 files) |
| `yarn build` | ✅ PASS |

## Issues

| # | Issue | Category | Severity | Location | Description | Steps to reproduce | Suggested fix |
| 1 | Auto-refresh polls the whole cell after a single-statement run | Query execution & data integrity | Minor | in-diff — `cellRefreshEngine.ts:636`, `:1198` | The refresh engine keys an entry on the full cell text (`sql: cell.value`), and `runGridRound` launches every statement in `entry.state.queries`. A "run single" or selection run does not narrow that set. One interval later the cell executes statements the user did not ask for, adds their tabs, and repeats them each tick. Rows stay correct and fresh. Writes stay blocked — a sibling DDL/DML still stops the whole cell. The cost is unexpected database load, not wrong data. | • Create a grid cell: `SELECT * FROM a; SELECT * FROM b;`. • Turn auto-refresh on (notebook default or cell override). • Put the cursor in statement 1 and run single. Only tab 1 appears. • Wait one interval. Both statements run, tab 2 appears, and both keep polling. | Decide one of two: scope a round to statements that already hold a result slot in `cell.result`, or document that auto-refresh always covers the whole cell. |
| 2 | Grid snapshots re-serialize an unchanged dataset every 10s | Performance & rendering at scale | Minor | in-diff — `cellRefreshEngine.ts:1358` (`persistGridFrame`) | Every settled grid round queues a snapshot. Once per 10s window `persistSnapshot` → `capResultBytes` runs a full `JSON.stringify` over the dataset (up to 10 000 rows) and writes it to IndexedDB, even when the frame did not change. The chart path skips this through the `entry.persistedResults` WeakMap; the grid path has no equivalent. **Verified as deliberate:** the comment above `persistGridFrame` states the re-write keeps `savedAt` current, so a reload does not show minutes-old data under a live poller. A `persistedResults` guard would break that intent. Cost is bounded by the per-entry 10s throttle and by visibility, but many large polling grids produce a steady stream of full re-serializes. | • Open a grid cell with ~10 000 static rows and auto-refresh `1s`. • Every 10s the dataset is re-serialized and written to IndexedDB with no content change. | Keep the current behavior, or write only a `savedAt` touch for an unchanged already-persisted frame so freshness survives without the re-serialize. |
| 3 | Agent-facing refresh state has no test through the AI/MCP snapshot | Test review & coverage | Moderate | out-of-diff (test gap) — `utils/ai/notebookSnapshot.ts` (`refreshFields` / `summarizeCells` / `serializeCell`) | No test populates a refresh map, so the wire mapping of `refreshing`, `last_refresh_error`, and `auto_refresh_blocked` is unexercised. A wrong wire key, a dropped field, or lost truncation would ship silently. The agent would then never learn that a cell's refresh fails or that a grid is write-blocked. That is a data-trust surface for agent flows. | • grep confirms no test builds a populated `readRefreshState()` and asserts the serialized cell or snapshot output. | Add a `notebookSnapshot` test. Use a controller whose `readRefreshState()` returns `lastRefreshError` plus `autoRefreshBlocked: "contains_write"`. Assert the wire cell carries `last_refresh_error` (truncated) and `auto_refresh_blocked`, and that `formatSnapshot` emits both. |
| 4 | Three load-bearing invariants have no asserting test | Test review & coverage | Minor | out-of-diff (test gap) — `NotebookProvider.tsx:805`/`:819`, `requestLimiter.ts:54`, `importTabs.ts:312` | (a) `noteRunStarted` / `noteRunFinished` must stay balanced when `runCell` or `reRunResultAt` is denied or throws. One leak disables that cell's refresh permanently. (b) `requestLimiter` must release its permit when the task rejects. A leaked permit slowly starves the shared 8-slot pool. (c) `autoRefreshMigrated` must survive `sanitizeBuffer`. If it regresses, tab copy and import re-stamp user-reset "inherit" cells back to Auto. The balancing `.finally`, the `finally release()`, and the round-trip line all exist. None has a test. | • Code inspection only. No runtime repro today. | Add three tests: a provider test that a rejected `runCell` leaves the engine ungated; a limiter test that a rejecting task releases and the queued task runs, in FIFO order; an `importTabs` test that `autoRefreshMigrated: true` round-trips. |
| 5 | "Reset cell overrides" counts with one metric and clears with another | State & context architecture | Minor | in-diff — `NotebookRefreshControl.tsx:54`, `:143` vs `NotebookProvider.tsx:781` | The menu item appears and labels itself from `countActiveAutoRefreshOverrides` (override **and** a visible view). `resetAutoRefreshOverrides` guards, reports telemetry, and clears with `countAutoRefreshOverrides` (all overrides). A cell that holds an override but shows no view is cleared without being counted. If every override sits on a view-less cell, the item never appears at all. **Verified as deliberate:** `notebookUtils.test.ts:2665-2700` names and tests the split — a dormant override sits on a cell that shows nothing, so it polls nothing. The residue is a labelling mismatch, not a stuck poll. | • Set a per-cell override, then clear that cell's result — or set the override through MCP on a result-less run cell. • Open the notebook refresh menu. The subtitle undercounts, or the item is missing while a dormant override is still stored. | Use one metric for both the label and the action, or leave as is and treat the reset as an intentional clean sweep. |
| 6 | The notebook write-blocked hint is invisible to screen readers | Accessibility & UX | Minor | in-diff — `NotebookRefreshControl.tsx:159` (`MenuHint`) | The hint "N cells contain DDL/DML and are excluded from auto-refresh" is a bare `div` inside a Radix `role="menu"`. Assistive technology in menu or application mode reads menu items only, so screen-reader users never hear the aggregate note. Sighted users see it. Keyboard navigation still works — roving tabindex skips the node. Per-cell write-blocked buttons carry their own `aria-label`, so the information is not lost entirely. | • Turn on a screen reader. • Put DDL/DML in one cell and open the notebook auto-refresh dropdown. • The aggregate hint is not announced. | Wrap the hint in `role="group"` with an `aria-label`, or fold the count into the interval trigger's `aria-label`, which already carries the override count. |
| 7 | Duplicated classify-or-fail preamble in the two fetch paths | Code structure, readability & types | Minor | in-diff — `cellRefreshEngine.ts:1070-1081` and `:1171-1182` | About 12 byte-identical lines repeat in `runChartFetch` and `runGridRound`: validate, catch, `setState({ classifyBlock: "failed" })`, return. That exceeds the 3-5 line duplication the team accepts. The offender and write-block handling below the preamble genuinely differs between the two paths and must stay separate. | • Read both fetch paths side by side. | Extract `classifyRoundOrFail(entry, round): Promise<ClassifiedStatement[] \| null>` that returns `null` on abort or failure. Keep the divergent write-block handling inline. |
| 8 | "Running…" feedback can lag when the limiter is saturated | Async / adversarial | Minor | in-diff — `useCellExecution.ts:586` | A cold multi-statement manual run classifies through the shared 8-slot `statementRequestLimiter` **before** it writes the initial `running` frame. Heavy background auto-refresh can hold every slot, so the placeholder appears only after a slot frees. No data issue and no deadlock. The DQL/DDL class cache limits the exposure to the first run of unseen statements, and single-statement runs skip the barrier. | • Leave many grids auto-refreshing so the limiter saturates. • Run a multi-statement cell whose statements were never classified. • "Running…" appears late. | Optional. Give run-barrier validation priority over background poll validation, or write the placeholder frame before the barrier resolves. |
| 9 | `noteRunStarted` runs before synchronous prep, outside any try | Async, timers & cancellation | Minor | in-diff — `NotebookProvider.tsx:805-819` | The `reRunResultAt` wrapper calls `engine.noteRunStarted(cellId)`, then runs `statementKeysFor(...)`, then returns the promise chain whose `.finally` balances the count. A throw in that synchronous prep would leak the count and disable the cell's refresh forever. Not reachable today — `statementKeysFor` over a `string[]` does not throw. This is hardening. | • Code inspection only. No runtime repro today. | Move `noteRunStarted` so it immediately precedes the returned `reRunResultAt(...)` call, or wrap the prep so a throw still reaches `noteRunFinished`. |

Verified but not tabled, all trivial: `statementKeysFor` is O(S²) per grid round and `classifyForRound` re-parses the cell SQL each tick — both bounded by the small statement count per cell (Agent 6 #2/#3); `MenuHint` uses a font size outside the menu subtitle scale (Agent 7 #2); choosing "Off" while the default is unset writes an explicit `false` and triggers a redundant resync (Agent 4 Nit A); the snapshot-persistence cluster is the one clean seam if the ~1600-line engine class ever needs a split (Agent 8 #2).

## False positives

| Category | Finding as reported | Why it is a false positive |
|---|---|---|
| Fresh-context adversarial (Agent 13, **primary**) | "Concurrent per-tab reruns strand a tab in a permanent Running… state." A second per-tab rerun on a different tab bumps the shared run generation, discards the first rerun's result, and leaves its tab on the running placeholder. | **Not reachable.** Only the ResultActionsBar rerun button calls `reRunResultAt`, and it is `disabled={isRunning \|\| !onReRun}` (`ResultActionsBar.tsx:139`). The first rerun adds the cell to `runningCellIds` **synchronously**, before its first `await` (`useCellExecution.ts:775`), and `isRunning={runningCellIds.has(cell.id)}` disables the button on **every** tab (`index.tsx:368`, `:656`, `:813`). A second concurrent `reRunResultAt` cannot start. The claim that tab B keeps a live rerun button is wrong. The other run paths (`runCell` via Run or Cmd+Enter) abort all controllers and rewrite the whole frame, so an overlapping run overwrites a placeholder instead of stranding it. The generation discard is defensive and correct. |
| Async, timers & cancellation (Agent 3) | "Manual per-cell refresh is silently dropped when it overlaps an in-flight run." | By design, and the effect is negligible. The `isRunPending \|\| isCellRunning` guard in `fetchOnce` (`cellRefreshEngine.ts:1002-1008`) lets the run win on purpose, and that run produces the fresh frame the click wanted. The adjacent comment states the intent. |
| Query execution & data integrity (Agent 1) | Considered: `resultsEquivalent` dedup could show stale rows; StatementKey collisions could mis-tab data; parallel execution could cross-contaminate results. | Ruled out against source. The dedup always walks the full per-cell set and only fast-rejects, so it cannot return a false "unchanged". `normalizeQueryText` plus an occurrence index makes keys distinct, so different SQL cannot collide. Parallel reads stay index-aligned, abort per query, and commit under a generation guard. |
| Persistence & migrations (Agent 5) | Considered: the migration is not applied on the live path; the marker is lost; snapshot fields break older readers. | Ruled out. The live mount (`beginNotebookMount` → `readNotebookView`), the preview path, and every agent read go through `migratePersistedNotebookView` (`notebookDexieView.ts:30`, `index.tsx:879`). The marker and the migrated cells are written together. New settings and snapshot fields are optional and non-indexed, with no DB version bump, so older readers ignore them. |

## Summary

**Verdict: Approve.** The remaining suggestions are minor and non-blocking.

This is a large, high-stakes change to query execution, result rendering, and persistence. It is unusually careful. All four quality gates pass on `51e1fc78`. The two highest-risk concerns going in were a write executing through auto-refresh, and legacy dashboards silently ceasing to poll after the migration. Both are **verified safe**: the refresh engine re-classifies fail-closed before every execution, so no cache-poisoning path exists, and the migration runs on the live path and stays backward compatible. Cross-context callers of every changed contract — statement-keyed results, `setActiveStatement`, `readRefreshState`, `setCellRefresh(undefined)`, the frozen grid timestamp — were walked two levels deep with **no broken call sites**.

- **Regressions and tradeoffs:** none that block. The one behavioral tradeoff is #1: auto-refresh covers the whole cell after a single-statement run. Data stays correct; unrequested statements re-execute on a loop. The efficiency tradeoff #2 is a documented choice that trades CPU for a current `savedAt`. Neither loses nor misrepresents data.
- **Findings:** 9 verified. One is Moderate (#3, a test gap); the rest are Minor. Four draft findings were dropped as false positives, most importantly Agent 13's "stranded Running… tab", which the `disabled={isRunning}` guard makes unreachable.
- **In-diff vs out-of-diff:** 7 in-diff, 2 out-of-diff. Both out-of-diff findings are test gaps (#3 and #4). The cross-context pass (Agent 12) ran to completion and cleared every call site, so the low out-of-diff count reflects genuinely clean consumers, not an under-run.

The strongest recommendation before merge is #3 — add the `readRefreshState` wire-mapping test, because it guards a surface where the AI can silently mis-report a cell's health. Then make a decision on #1: scope auto-refresh to the statements the user ran, or document the full-cell behavior.

bluestreak01 (approved):

PR #598, opened 2026-08-18
refactor: remove autoRefreshMigrated marker and implicit chart auto-refresh migration

emrberk:
## What

Removes the `autoRefreshMigrated` marker field and the `migrateImplicitChartAutoRefresh` load-time migration from #595. `undefined` auto-refresh on a chart cell now always means "inherit the notebook default", with no special case for records saved before #595.

## Why

The marker served one very narrow purpose: it made a one-time stamp idempotent. The stamp itself only converted an implicit legacy behavior into an explicit override. That is a dummy field from the user's point of view — no one sets it, no one reads it after the first load, yet every notebook record would carry it forever. Removing the migration keeps the settings shape minimal and the read path free of version-detection logic.

## Downside (accepted)

Charts saved before #595 polled through an implicit per-view Auto. After this change they load as "inherit", which resolves to Off when the notebook has no default:

- They freeze silently. A frozen chart shows its last snapshot and looks identical to a live chart.
- Recovery is manual, per cell. Setting the notebook default to Auto overshoots: grids inherit it too and start polling, which they never did before.
- The change is one-way. Once new code reads an old record, legacy `undefined` and deliberate "inherit" become indistinguishable, so this migration cannot be shipped retroactively.

#595 has not been in a tagged release, so the exposure today is dev environments. Production users get the new semantics directly when the next release ships. Browsers that already ran a stamped build keep an inert `autoRefreshMigrated: true` property in stored records; nothing reads it, and Dexie needs no schema change.

### DIFF
--- a/src/scenes/Editor/Monaco/importTabs.ts
+++ b/src/scenes/Editor/Monaco/importTabs.ts
@@ -309,7 +309,6 @@ const sanitizeNotebookSettings = (
   }
   if (isAutoRefresh(item.autoRefreshDefault))
     settings.autoRefreshDefault = item.autoRefreshDefault
-  if (item.autoRefreshMigrated === true) settings.autoRefreshMigrated = true
   return settings
 }
 
--- a/src/store/notebook.test.ts
+++ b/src/store/notebook.test.ts
@@ -2,9 +2,7 @@ import { describe, it, expect } from "vitest"
 import {
   dropLegacyChartConfigs,
   migrateCellName,
-  migrateImplicitChartAutoRefresh,
   migrateLegacyCellNames,
-  type AutoRefresh,
   type NotebookCell,
   type NotebookViewState,
 } from "./notebook"
@@ -14,13 +12,6 @@ const cell = (over: Partial<NotebookCell> & { id: string }): NotebookCell => ({
   value: "",
   ...over,
 })
-
-const chart = (id: string, autoRefresh?: AutoRefresh): NotebookCell =>
-  cell(
-    autoRefresh === undefined
-      ? { id, mode: "draw" }
-      : { id, mode: "draw", autoRefresh },
-  )
 
 describe("migrateCellName", () => {
   it("promotes a legacy chartConfig.name to the cell name and drops the old copy", () => {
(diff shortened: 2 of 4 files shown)


emrberk:
# PR #598 Review — `refactor: remove autoRefreshMigrated marker and implicit chart auto-refresh migration`

Reviewed at level 1 (semantic delta + 4 agents + full quality gate). Head commit `a14d2cd3`.

## Issues

| # | Issue | Category | Severity | Location | Description | Steps to reproduce | Suggested fix |
| 1 | Legacy charts freeze on stale data | Query execution & data integrity | Accepted (no action) | in-diff (`src/utils/notebooks/notebookDexieView.ts:30`) | Shipped 2.0.2 used `cell.autoRefresh ?? true` for draw cells (`chartRefreshEngine.ts:319,342` at commit `1ec322c5`), and the 2.0.2 dropdown had no "inherit" item — so `autoRefresh: undefined` in stored 2.0.2 data means exactly "user never touched it, chart was live". After this PR it resolves through `resolveAutoRefresh` (`notebookUtils.ts:59`) to `false`. Worse than "stops polling": `ensureData` (`cellRefreshEngine.ts:844-856`) sees a settled snapshot with rows, sets `lastFetchedAt` from `snapshot.savedAt` and **returns without fetching**. The chart paints the previous session's rows forever. There is no passive freshness signal: `CellBottomContent.tsx:89-98` passes `timestamp` only on the grid branch, and the `Off` label lives in `CellWideActions`, rendered only at `toolbarTier === "expanded"` (`Cell.tsx:442,464`) — narrow grid/dashboard cells show nothing. State is reachable only by opening the kebab menu (`CellToolbar.tsx:386`). | - On 2.0.2, create a notebook, draw a chart, never touch auto-refresh<br>- Upgrade to a build containing this PR<br>- Reload the console and open the notebook<br>- The chart renders rows from the earlier session and never refetches; in grid layout no control indicates it is Off | None for this PR — the tradeoff is the decision this PR makes. Optional follow-up: render a passive stale/Off marker on the chart itself (a `savedAt` badge) that survives the `standard`/`compact` tiers |
| 2 | PR body understates production exposure | PR description | Minor | in-diff (description) | The body says "#595 has not been in a tagged release, so the exposure today is dev environments." That is true only for the `autoRefreshMigrated` marker. The freeze in #1 hits every shipped-2.0.2 user's existing charts the moment the next release ships. The following sentence concedes this, but the leading claim reads as "dev-only". The "Why" section is also written in internals language (marker field, read path, version-detection) rather than user impact. | - Read the PR body top-down<br>- Conclude the freeze is a dev-only concern | Reword to "the marker exposure is dev-only; the chart freeze lands on every 2.0.2 user at the next release" |
| 3 | New invariant has no test | Test review & coverage | Minor | in-diff (`src/store/notebook.test.ts`) | The PR deletes 5 tests and adds none. `migratePersistedNotebookView` has zero direct tests, so nothing pins the invariant this PR establishes — that a persisted draw cell with `autoRefresh: undefined` comes back untouched. A future re-introduction of a load-time stamp would be caught by no test. Separately, `cellRefreshEngine.test.ts:60-69` types `drawCell(id, value, autoRefresh)` with `autoRefresh` required, so the engine is never exercised with `undefined` on a draw cell. (`resolveAutoRefresh` itself is covered at `notebookUtils.test.ts:2654-2662`.) | n/a — coverage gap | Add one test asserting `migratePersistedNotebookView` leaves `cell.autoRefresh` absent on a draw cell with no notebook default; make `autoRefresh` optional in the `drawCell` helper and add an inherit case |
| 4 | Dead key persists forever | Query execution & data integrity | Minor | out-of-diff (`src/scenes/Editor/Notebook/notebookUtils.ts:500-510`) | Contract violated: `NotebookSettings` no longer declares `autoRefreshMigrated`, but `buildPersistPayload` writes `settings` verbatim and every transition spreads `{ ...parts.settings }`. A record stamped by a #595 dev build round-trips the undeclared key through Dexie indefinitely; TS does not catch it (spreads bypass excess-property checks). The PR calls this "inert" — correct, nothing reads it, and import strips it (`importTabs.ts` whitelist). Cosmetic, dev DBs only. | - Run a #595 dev build, open a notebook (stamps the marker)<br>- Update to this PR<br>- Inspect the buffer row in IndexedDB — `autoRefreshMigrated: true` is still there | Leave as-is, or strip the key once in `migratePersistedNotebookView` |

## False-positives

| Category | Description | Explanation |
|---|---|---|
| Query execution & data integrity | `settings` becoming possibly `undefined` after migration breaks consumers | Every consumer already normalizes: `partsOf` (`notebookDexieView.ts:112`), `notebookSnapshot.ts:212`, `notebookHeadlessRun.ts:332`, `NotebookProvider.tsx:241`. Typecheck clean. |
| React correctness & hooks | Losing the always-fresh object identity breaks a memo/effect/dirty check | No caller compares the view by reference. `previewSeed` (`index.tsx:875-881`) already keyed on `previewView`. `NotebookProvider` reads `initialState` only in mount-time initializers. Aliasing is safe — every writer copies (`useCellsStore.ts:26-56`, `notebookUtils.ts:609,662,961-971`, `NotebookProvider.tsx:387`), and preview never persists. |
| Async, timers & cancellation | Off resolution leaves a dangling poll or a state update after unmount | The Off path allocates nothing: `updatePoll` (`cellRefreshEngine.ts:934-946`) sets `pollKey = null` and aborts. No `setInterval` exists; cadence is `sleep(ms, signal)`. `destroy()` -> `removeEntry` aborts every round. |
| React correctness & hooks | Stale closure over `autoRefreshDefault` / `cell.autoRefresh` | `this.autoRefreshDefault` is assigned before the `sameEntryCells` early return (`cellRefreshEngine.ts:328`); `runPollLoop` re-reads per iteration; UI resolves per render from props. |
| Async, timers & cancellation | Legacy charts render empty because nothing fetches | `ensureData` falls through to `void this.fetchOnce(entry)` (`cellRefreshEngine.ts:869`) when no snapshot exists, and `refreshAll` ignores `autoRefresh`. Never-empty holds — the problem is staleness (#1), not blankness. |
| Async, timers & cancellation | First fetch loses its random stagger for legacy charts | Bounded by the global `createRequestLimiter(8)`, and it is the same path every born-Off chart already takes. Not a regression from this PR. |
| Test review & coverage | The `e2e/questdb` submodule bump is unrelated and should be split out | `git log -- e2e/questdb` shows the pointer moves on nearly every commit on `main`. Repo norm, not PR noise. |
| Test review & coverage | MCP / shared-definitions contract drift | `src/consts/shared-definitions.json` never enumerated `autoRefreshMigrated`; `applyNotebookState.ts:376-378` maps only `layoutMode` and `autoRefreshDefault`. The byte-identical sync test cannot drift. |

## Summary

**Approve.** The code is correct and the behavior change is the deliberate product decision this PR makes.

- Quality gate on the real PR head `a14d2cd3`: `yarn typecheck` PASS, `yarn lint` PASS, `yarn build` PASS, `yarn test:unit` PASS (1973 tests, 87 files). No rows added to the table. (Note: an agent reported 3 `Metrics/utils.test.ts` failures — an artifact of running bare `npx vitest run`; `test:unit` sets `TZ=UTC` and they pass.)
- The refactor is clean and complete: no dangling references, no broken consumers, no leaked timers, no identity or nullability regressions.
- Issue #1 is the accepted tradeoff, recorded here only so the shipping behavior is documented: the chart does not merely stop polling, it renders the previous session's snapshot and never refetches, with no passive signal in grid layout. Nothing to fix in this PR.
- Remaining findings are Minor: a PR-description wording nit and a test-coverage gap.
- 4 findings verified (1 accepted by design, 3 Minor), 8 dropped as false positives. All 4 in-diff except #4 (out-of-diff, `notebookUtils.ts` persist path). Agent 12 does not run at level 1, so a low out-of-diff count is expected.

bluestreak01 (dismissed):

bluestreak01 (approved):

PR #600, opened 2026-08-19
feat: add light/dark theme support, shared component style updates

emrberk:
## Summary

Adds full light/dark theme support across the Web Console, including system-theme detection, persisted user preference, Monaco editor themes, charts, grids, dialogs, menus, and shared components.

## Highlights

- Introduces semantic design tokens and removes legacy hard-coded color usage.
- Adds a theme selector with System, Light, and Dark modes.
- Updates shared buttons, controls, menus, icons, focus states, and component styling.
- Replaces the default Result Quick Vis experience with a configurable ECharts chart panel.
- Adds chart type, axis, series, partitioning, zoom, and OHLC controls.
- Retains the legacy Quick Vis panel behind `?useQuickVis=1`.
- Refreshes notebook, AI assistant, schema, import, search, footer, and onboarding surfaces.
- Preserves existing preferences and persisted metric colors during migration.


emrberk:
# Code review — PR #600

**Title:** `feat: add light/dark theme support, shared component style updates`
**Base:** `dc6b8e45741e1166f90d1f5a8858bec3bd779a93` · **Head:** `7fd91462da23a8ff100fac44ed56eec04bb71100` (branch `feat/themes`)
**Size:** 314 files, +11,991 / −7,868
**Review level:** 3 (full mission-critical pass)

---

## PR title and description

The title follows Conventional Commits. The description does not repeat the verb and speaks to
end-user impact rather than internals. Two items are missing:

- The description does not mention that chart timestamps now render in UTC instead of browser-local
  time. This affects existing notebook charts, not only the new panel.
- "Retains the legacy Quick Vis panel behind `?useQuickVis=1`" is accurate, but that panel received
  260 new lines of SCSS and lost its only e2e test. See issue #6.

---

## Quality checks

| Check | Command | Result |
| ----- | ------- | ------ |
| Type errors | `yarn typecheck` | Pass |
| Lint | `yarn lint` (incl. new `lint:colors`) | Pass |
| Unit tests | `yarn test:unit` | Pass — 88 files, 1992 tests |
| Build | `yarn build` | Pass |

No quality-check failures. All four ran at head `7fd9146`.

---

## Issues

| ID | Issue name | Category | Severity | Location | Net impact | Evidence |
| -- | ---------- | -------- | -------- | -------- | ---------- | -------- |
| #1 | Non-chartable result draws empty chart | Query execution & data integrity | Moderate | in-diff | Any user charting UUID/IPv4/LONG256 results; every such query | Scratch vitest at head — see detail |
| #2 | Chart total contradicts toolbar badge | Query execution & data integrity | Moderate | in-diff | Users charting live-ingest queries; every chart view | N/A — static, `Result/index.tsx:399`, `ResultChart/index.tsx:176` |
| #3 | Theme switch rebuilds every chart | React correctness & hooks | Moderate | in-diff | Chart users who switch theme or use `system`; per switch | N/A — static, `ChartRenderer.tsx:224,199-212`, `buildEchartsOption.ts:547` |
| #4 | Light mode toggle labels fail AA | Accessibility & UX | Moderate | in-diff | Low-vision users in light mode; every notebook cell | Computed contrast from `src/theme/index.ts` at head |
| #5 | Metric colour write-back is gratuitous | Persistence & migrations | Moderate | in-diff | Every user with saved metrics; one permanent IndexedDB rewrite | Scratch vitest at head + read-site inventory |
| #6 | Migration safety nets not wired up | Test review & coverage | Moderate | in-diff | The whole token migration; silent erosion over time | N/A — static, `ci.yml`, `package.json:188`, `e2e/commands.js:547` |
| #7 | Per-cell glass layers survive virtualization | Performance & rendering at scale | Moderate | in-diff | Large-notebook users; scales with cell count, bound 200 | N/A — static skeleton; paint cost not measured |
| #8 | Duplicated new shared-component code | Code structure, readability & types | Moderate | in-diff | Developers; every future edit to these primitives | N/A — static, four cited duplications |
| #9 | Assorted standards and ARIA nits | Code structure / Accessibility | Minor | in-diff | Developers, plus screen-reader users on three tab strips | N/A — static, per-item citations |

### #1 — Non-chartable result draws an empty chart instead of the guard message

**Evidence.** Scratch vitest run at head (not committed): `isResultChartable([UUID, IPV4])` returns
`false`, but `inferChartConfig(...).chart.yColumns` returns `["ip"]`. Running the full ResultChart
pipeline — `resolveDraw` then `buildEchartsOption` — produces `xAxis.type === "category"` with every
raw UUID string as a label, and `series[0].data === [null, null, null]`.

**Description.** `ResultChart` computes `hasChartableColumns` from `yColumns.length > 0`
(`ResultChart/index.tsx:171-175`). The final fallback in `inferChartConfig`
(`inferChartConfig.ts:197-199`) returns `columns[0]` plus `columns.slice(1, 11)` whatever the column
types are. Therefore `yColumns` is never empty for a result with two or more columns. The message
"This result has no chartable columns." only renders for a single-column result. The repository
already contains the correct predicate, `isResultChartable`, and no production code calls it. The
user sees a category axis carrying up to 10,000 raw values and no plotted marks.

**Steps to reproduce.**

- Run `SELECT id, ip FROM logs` (UUID + IPV4). Any two columns that are not temporal, numeric or
  categorical will do — two LONG256 columns, or two ARRAY columns.
- Switch the result panel to Chart.
- The chart area shows a dense category axis and no bars.

**Suggested fix.** Gate `canDraw` on `isResultChartable(chartResult.columns)` in
`ResultChart/index.tsx`. As an alternative, make the `inferChartConfig` fallback return empty
`yColumns` when `groupColumns` places every column in `other`.

### #2 — The chart can show a row total that contradicts the toolbar badge

**Evidence.** Static. `Result/index.tsx:399-408` renders `count` from the grid execution, and that
badge sits outside the `viewMode === "grid"` guard. `ResultChart/index.tsx:176-178` renders
`chartResult.count`, which comes from the separate `src:"vis"` execution in
`useChartQuery.ts:106-111`.

**Description.** The chart re-runs the SQL as its own execution. On a table under ILP ingestion the
two executions return different totals. Both numbers render at the same moment: the toolbar badge
from execution 1, the truncation banner from execution 2. Nothing tells the user that two executions
occurred. At base, Quick Vis also re-ran the query, but it sent `count: false` and had no truncation
banner, so two conflicting totals could not appear together.

**Steps to reproduce.**

- Run `SELECT * FROM trades` against a table that is receiving ILP writes.
- Switch the result panel to Chart and wait for the fetch to settle.
- Compare the toolbar row-count badge with the "First N of M rows are shown" banner.

**Suggested fix.** Label the banner as the chart's own execution. As an alternative, take the total
from the grid's `count` and use the chart fetch only for the plotted rows.

### #3 — A theme change tears down every chart, losing legend state and replaying entry animation

**Evidence.** Static. `ChartRenderer.tsx:224` sets ``key={`${theme.mode}:${key}`}``.
`buildEchartsOption.ts:547` sets `legend: baseLegend` with no `selected` map, and nothing anywhere
writes `legend.selected`. The `events` memo at `ChartRenderer.tsx:199-212` wires only `datazoom`.
`firstInstanceDoneRef` is set at `ChartRenderer.tsx:167` and is never reset.

**Description.** Changing the key disposes the ECharts instance and initialises a new one. The code
restores zoom from `zoomWindowRef`, but it does not restore legend series-visibility. Every series
the user hid reappears. Separately, `suppressEntryAnimation` evaluates
`!animateEntryRef.current && !firstInstanceDoneRef.current`, so after the first instance the Result
panel replays its entry animation even though its caller passes `animateEntry={false}`. With the
`system` preference, an OS appearance change triggers both effects with no user action.

**Steps to reproduce.**

- Draw a chart with two or more series, in a notebook cell or in the Result chart panel.
- Click a legend entry to hide one series.
- Switch the theme in the editor menu.
- The hidden series returns, and the chart animates from zero.

**Suggested fix.** Capture `legendselectchanged` into a ref and pass it back as `legend.selected` on
re-init, next to the existing zoom restore. For the animation, read the `animateEntry` prop on each
instance instead of latching `firstInstanceDoneRef`.

### #4 — Light mode: notebook Run/Draw toggle labels fail WCAG AA

**Evidence.** Contrast computed from `src/theme/index.ts` at head. In light mode,
`statusInfo #176f87` on `controlTrack #d9dce2` gives **4.18:1**, and `statusSuccess #067047` on the
same track gives **4.47:1**. The AA threshold for normal text is 4.5:1. The dark equivalents on
`controlTrack #262833` give 8.80:1 and 6.19:1.

**Description.** `CellRunDrawToggles.tsx:64,96` pass `$size="md"`, which resolves to
`theme.fontSize.md` = 1.4rem. The root font size is 10px, so the label renders at 14px, weight 500.
That is normal text, so the 4.5:1 threshold applies. The active state does not correct this, because
the tone colour is the inactive colour. The hover state composites `interactionHover` onto the
track, which lightens it and makes the ratio slightly worse.

**Steps to reproduce.**

- Set the theme to Light.
- Open a notebook and widen a cell until the toolbar shows text labels.
- Read the inactive "Draw" label against the toggle track.

**Suggested fix.** Darken the light-mode `statusInfo` and `statusSuccess` values. As an alternative,
give the segmented control a dedicated inactive-label token that meets 4.5:1 on `controlTrack`.

### #5 — The metric colour write-back rewrites persisted data for no benefit

**Evidence.** Scratch vitest run at head confirms the migration itself is lossless: all ten legacy
persisted hexes map 1:1 onto `dataSeries1..10` and resolve to identical dark-mode colours. A read-site
inventory shows every consumer already normalises: `metric.tsx:97`, `add-metric-dialog.tsx:184-185`,
and `importTabs.ts:183` all call `toMetricColorToken`. The only other use of the raw value is the
write-back itself at `Metrics/index.tsx:192-202`.

**Description.** Because every read site already normalises at read time, the effect changes nothing
on screen. Its only effect is to replace a `#RRGGBB` literal in IndexedDB with a token string such
as `"dataSeries3"`. A console build from before this PR then passes that token to uPlot as a CSS
colour, which the canvas rejects. `importTabs` at that version fails its hex test and substitutes
`#FF6B6B` for every metric. The write puts durable user state at risk and returns nothing.

**Steps to reproduce.**

- Open a Metrics tab that already holds widgets, on this build.
- Inspect `db.buffers` — `metricsViewState.metrics[].color` now holds tokens, not hexes.
- Export tabs, then import that file into a console built before this PR. Every metric renders red.

**Suggested fix.** Delete the effect at `Metrics/index.tsx:192-202`. Read-time normalisation already
covers every consumer.

### #6 — The migration's two safety nets are not wired up

**Evidence.** Static. `.github/workflows/ci.yml` runs build, bundlewatch, `test:unit`, and the two
e2e suites. It never runs `yarn lint` or `yarn typecheck`. `lint-staged` (`package.json:188-192`)
runs eslint on `*.{ts,tsx}` only. `e2e/commands.js:547` sets
`appearance.themePreference = "dark"` in the shared setup command. No spec references
`theme-mode-trigger`, `theme-mode-light`, or `theme-mode-dark`. No unit test imports `lightTheme` or
`lightColors`.

**Description.** Two gaps compound each other. The new `check-color-literals.mjs` guard exists to
protect the token migration, but it runs only when a developer types `yarn lint` by hand. Separately,
the light palette — the headline feature of this PR — has no automated coverage at all, because the
harness pins every Cypress spec to dark. A regression in light mode ships with no signal. A related
change at `e2e/commands.js:320` broadened `getEditor()` from `.monaco-editor.vs-dark` to
`.monaco-editor`, which removed the suite's only observation of Monaco's applied theme.

**Steps to reproduce.**

- Add a hard-coded hex to any `src/**/*.tsx` file and commit it. The pre-commit hook and CI both pass.
- Break a light-mode-only token value. Every Cypress spec still passes.

**Suggested fix.**

- Add a `yarn lint` step to `ci.yml`.
- Add `"*.{ts,tsx,scss,css}": ["node scripts/check-color-literals.mjs"]` to `lint-staged`.
- Add one Cypress spec that seeds `appearance.themePreference = "light"`, drives the `theme-mode-*`
  hooks, and asserts that `.monaco-editor`'s background differs between the two modes.

### #7 — Every notebook cell keeps a promoted, backdrop-filtered layer while off screen

**Evidence.** Static skeleton only. `SegmentedControl/index.tsx:86-89` sets `backdrop-filter` and
`will-change: transform, width`. `SegmentedControl/index.tsx:261-300` mounts a per-instance SVG
`feDisplacementMap` filter. `Cell.tsx:405-425` swaps only the cell body for a placeholder when
`contentMode === "placeholder"`; `CellDragHeader` renders unconditionally and carries
`CellWideActions` → `CellViewToggle` → `SegmentedControl`. The base `CellViewToggle` was a plain
styled div with no filter and no `will-change`. `MAX_NOTEBOOK_CELLS` is 200
(`src/store/notebook.ts:9`).

**Description.** Content virtualization does not unmount the cell header, so off-screen cells keep
their glass layer. `will-change: transform` promotes a compositing layer whatever the opacity is.
The count of live layers therefore tracks total cell count, not what is on screen. **I could not
measure the paint cost in this environment.** The static fact is the finding; I am not asserting a
measured frame-rate regression.

**Steps to reproduce.**

- Open a notebook with many cells and run them.
- Scroll so most cells leave the viewport.
- Inspect the layer tree — each cell retains a promoted, backdrop-filtered element.

**Suggested fix.** Remove `will-change` — the animation already writes `transform` directly. As an
alternative, render `GlassSelection` and its SVG filter only while the cell is inside the
virtualization band.

### #8 — The shared-component layer picked up avoidable duplication

**Evidence.** Static. All four items are net-new in this PR.

**Description.**

- The liquid-glass selection animation exists twice, about 44 lines each:
  `SegmentedControl/index.tsx:190-233` and `ReactChromeTabs/chrome-tabs.ts:388-431`. Both use the
  same centre/travel maths, cubic ease-out, sine stretch envelope, rAF tick, and cancel bookkeeping.
  The copies already differ in easing exponent and stretch curve with no stated reason.
  `src/components/LiquidGlass/` exists and holds only `createLiquidLensMap.ts`.
- `SelectMenu/index.tsx:182-215` re-declares `ItemIcon`, `ItemCopy` and `ItemDescription`, which
  `menuStyles.ts:39-61` already exports. The same file already imports from `menuStyles`.
  `SelectMenu.Divider` at `:393-397` is identical to the `DropdownMenu` one.
- `MultiSelect/index.tsx:26-51,104-115` re-implements `SelectMenu`'s item scaffolding and its
  `modal={false}` default, while importing `SelectMenu` for the content surface.
- `Button` exposes both `prefixIcon` (about 20 callsites) and `leadingIcon` (6 callsites), resolved
  by `leadingIcon ?? prefixIcon` at `Button/index.tsx:126`. Neither carries a deprecation marker.
  The new `Button/README.md` documents the `skin` removal and the `BadgeType` alias, but never
  mentions this pair.

**Steps to reproduce.** Not applicable — maintainability, no runtime trigger.

**Suggested fix.** Extract the glass animation into `src/components/LiquidGlass/`. Reuse the
`menuStyles` leaf components in `SelectMenu`, and compose `MultiSelect` on top of `SelectMenu`.
Choose one icon prop name, codemod the remaining callsites, and record the choice in the README.

### #9 — Assorted standards and ARIA nits

**Evidence.** Static, with a citation per item.

**Description.**

- New `role="option"` rows at `QueryPicker/Row/index.tsx:88,94` have no `role="listbox"` ancestor.
- New `role="tablist"` / `role="tab"` strips at `TableDetailsDrawer/index.tsx:683-721`,
  `ChartSettingsDrawer.tsx:410-416` and `SettingsModal.tsx:958-965` have no `aria-controls`, no
  associated tabpanel, no roving tabindex, and no arrow-key handler. Base had no roles at all, so
  the change makes the declared semantics invalid rather than degrading a working flow.
- `ResultChart/index.tsx` breaks the member order in `CLAUDE.md`: an effect at `:180` precedes
  functions at `:186` and derived state at `:191`.
- Nine callsites pass `aria-pressed` that `SegmentedControlButton.attrs` overwrites
  (`SegmentedControl/index.tsx:316-320`): `SearchPanel.tsx:789,800,811`,
  `NotebookViewToggle.tsx:46,58`, `CellViewToggle.tsx:126,146`, `CellRunDrawToggles.tsx:65,97`.
- `NotebookViewToggle.tsx:12-14` re-sets `background: controlTrack`, the exact token its base already
  applies. Its comment claims to be the single surface definition; the real one is
  `SegmentedControl/index.tsx:29`.
- `createDraculaTheme`'s `mode` parameter defaults to `"dark"`, but both callers pass it explicitly.
- `ThemeModeProvider/index.tsx:61` mutates module state during render with no comment. The placement
  is defensible, because non-React renderers read the value in the same commit. That reasoning
  appears nowhere in the code.
- `quick-vis.ts:116` adds `addToBody: true` with no matching `destroy()`. Three `.ss-content` panels
  leak to `document.body` per Result remount. This reaches only the `?useQuickVis=1` path.
- `THEMING.md:52-53` reads "contentAccent in light mode and actionPrimaryHover in light mode". The
  first should read dark mode.

**Steps to reproduce.** Not applicable for the structural items. For the tab strips: open Schema, a
table, then the details drawer, and press Left or Right arrow. Nothing happens, though the role
announces that it should.

**Suggested fix.** Add `role="listbox"` to the QueryPicker wrapper, or drop the row roles. Give the
three tab strips `aria-controls`, a `role="tabpanel"` target, roving tabindex, and arrow-key
handling — or drop the roles. Take the other items individually; each is a one-line change.

---

These are pre-existing bugs on `$BASE`, found in code this review visited. This change does not
introduce, break, or worsen them. They do not affect the verdict.

**1. ChromeTabs never tears down**

- **Net impact:** all console users; leak bounded by tab-bar lifetime.
- **Location:** `src/components/ReactChromeTabs/chrome-tabs.ts:175-180`, `hooks.tsx:116-120`
- **Symptom:** the `ResizeObserver`, scroll and click listeners are never removed. This PR removes
  one leak here (the un-removable window resize listener) and adds two uncancelled rAF handles.
- **Reachability:** the tab-bar element does not unmount in practice, so no user sees a symptom today.
- **Suggested fix:** return a disposer from `ChromeTabs` that disconnects the observer, removes the
  listeners, and cancels both rAF handles. Call it from the `useChromeTabs` effect cleanup.
- **Severity if filed standalone:** Minor

**2. quick-vis leaves subscriptions behind**

- **Net impact:** users on `?useQuickVis=1`; grows per Result remount.
- **Location:** `src/js/console/quick-vis.ts` — `bind()`
- **Symptom:** subscriptions to `MSG_ACTIVE_SIDEBAR`, `MSG_QUERY_DATASET` and `$(window).resize` are
  never removed. The disposer added by this PR removes only `MSG_THEME_CHANGED`.
- **Reachability:** legacy path only, behind the opt-in query parameter.
- **Suggested fix:** unsubscribe all three in the returned disposer.
- **Severity if filed standalone:** Minor

**3. `setValue` has no error handling**

- **Net impact:** Safari private mode and quota-exceeded users; any settings write.
- **Location:** `src/utils/localStorage/index.ts:28`
- **Symptom:** `localStorage.setItem` throws out of the caller. Callers set React state before the
  write, so the UI shows a preference that was never persisted.
- **Reachability:** every settings write, including the new theme preference.
- **Suggested fix:** wrap the write in try/catch and surface a failure to the caller.
- **Severity if filed standalone:** Moderate

**4. Menu keyboard focus is nearly invisible**

- **Net impact:** keyboard and low-vision users; every menu.
- **Location:** `src/components/menuStyles.ts`
- **Symptom:** focus is shown with `interactionHover` alone — about 1.17:1 dark and 1.15:1 light,
  against a 3:1 threshold for non-text indicators. Base measured about 1.32:1, so both fail.
- **Reachability:** open any dropdown and navigate with the keyboard.
- **Suggested fix:** add the shared focus ring to menu items, as global styles already do elsewhere.
- **Severity if filed standalone:** Moderate

**5. `graph.tsx` binds one ref to two elements**

- **Net impact:** developers; no user-visible symptom found.
- **Location:** `src/scenes/Editor/Metrics/graph.tsx`
- **Symptom:** `graphRootRef` is attached to both `<Root>` and an inner `<div>`. The inner one wins
  on commit, so the outer binding is dead.
- **Reachability:** every Metrics render.
- **Suggested fix:** remove the binding that is not read.
- **Severity if filed standalone:** Minor

---

## Summary

**Verdict: approve with comments.**

I want #1 and #5 addressed before merge. #1 shows the user a broken chart where the component already
contains the correct message. #5 writes to durable user state for no gain, and the fix is a deletion.
The other seven items can follow.

**Correctness gate: passes.** No admitted Critical remains open.

**Test gate: passes.** Admitted Critical coverage gaps: 0. One Moderate coverage gap (#6). The light
theme has no automated coverage, but it is a new opt-in surface with no prior behaviour to regress,
so it does not fail the gate.

**In-diff / out-of-diff split:** 9 in-diff, 0 out-of-diff. This is genuine containment, not an
underrun of the cross-context pass. Evidence:

- The callsite pass walked 92 unchanged files that import a changed module and cleared each one.
- `skin` → `variant` left no residue anywhere in `src/` or `e2e/`.
- No `data-hook` used by e2e lost its source. `chart-panel-button` is generated by
  ``data-hook={`${mode}-panel-button`}`` at `Console/index.tsx:253`.
- All 57 SCSS `theme-color()` names, all `getThemeColor()` names, and all 29 Monaco tokens resolve in
  both palettes.
- `darkColors` and `lightColors` carry identical 139-key sets.
- Across `AIChatWindow`, `Import`, `Notebook/cells`, `SetupAIAssistant`, `Schema` and `Footer` —
  8,499 diff lines, 1,811 added — there are zero added `useState` or `useEffect` calls. Those
  directories are presentation-only.

**Severity distribution:** 0 Critical, 8 Moderate, 1 Minor.

**Regressions and tradeoffs.**

- `buildEchartsOption` now sets `useUTC: true`. Chart timestamps therefore render in UTC, not in the
  browser's zone. This is a correct fix, because it matches the result grid, and the new test at
  `buildEchartsOption.test.ts` covers it. It is a visible change for users outside UTC, and the PR
  description does not mention it.
- The chart fetch requests every column rather than only the charted ones. It is capped at 10,000
  rows, where Quick Vis had no cap. This is better for tall results and worse for short, wide ones.
  The settings panel needs the full column list to allow re-picking without a refetch, so the
  tradeoff is deliberate.
- Being on the Chart tab now runs a second execution of the query automatically, where Quick Vis
  required an explicit "Draw" click. The fetch carries its own Stop control, supersedes itself, and
  aborts when the tab loses visibility.

bluestreak01 (approved):

PR #606, opened 2026-09-01
refactor: redesign notebook cell dimensions and responsive views

emrberk:
### Requested: stable chart sizing across viewport changes

Reported problem: charts squashed and stretched while moving windows or opening the Editor panel, so the chart shape did not hold.

- Store the pane layout per cell (`paneView`) and make it independent of the viewport. Under 480 px the compact tier used to hide the editor and stretch the chart to the full cell height; widening collapsed it back. The tier now affects only the toolbar.
- Keep the result pane at its own height when the editor is hidden, instead of editor + result height, so toggling the editor never resizes the chart.
- Apply a grid south-edge resize to the result pane only. The old proportional scaling of editor and chart is gone.
- Give charts a 296 px floor and every pane a 2400 px ceiling. Drag, keyboard, and agent writes clamp to the same bounds.
- Suppress grid item transitions while the container width changes, and never animate item height, so cell boxes land in one step.
- Hide overlapping x-axis labels so a narrow chart thins them instead of colliding.
- Show the zoom slider and wheel zoom from the measured width and data density, with one option structure so a resize never remounts the chart.
  - Thresholds (per rendered mark, from plot width / slot count): slider when a bar or candle body is under 6 px, or a line/scatter point is under 1 px; wheel zoom at 3x those floors. A stack counts as one column; pie has no zoom.
- Keep Reset Zoom available when a resize drops the slider (covered by e2e).

Note: the chart width still follows the container. This PR stabilizes the height; it does not lock a width-to-height ratio.

### Other improvements

- Replace grid-height and maximized-view agent semantics with independent `editor_height`, `result_height`, and `view` controls. Snapshots report the stored view, mode, and semantic heights.
- Add a per-cell Stop control for cancelling initial query runs and chart fetches.
- Treat hiding a result as an explicit discard action that clears the current result and saved snapshot while preserving the preferred layout for future runs.
- Preserve run history and copy matching saved result snapshots when duplicating notebooks, without switching the active tab.
- Return faithful JSON previews and preserve semantic dimensions in snapshots.
- Add keyboard resizing to pane handles (arrow keys, Home, End) with ARIA value attributes.
- Compare statements by formatter identity so whitespace and casing edits keep the chart frame and run nothing.

## Validation

- yarn typecheck
- yarn test:unit (2,160 tests across 95 files)
- focused notebook snapshot/tool/controller tests (252 tests)
- yarn build
- targeted ESLint and git diff --check

## Deferred follow-up

- Update mcp-server-questdb for the new notebook tool contract.
- Bump the MCP server version.
- Update EXPECTED_MCP_VERSION in this repository after that release.


emrberk:
# PR #606 Review

Reviewing PR #606 at level 3

Reviewed head: `a36aa4e812b15270f237aaddb27ab6f3d2031d34`  
Base revision: `72ca5487b263fba39ecd2e9221066b03535b4c75`

The title and description follow the repository conventions.

## Issues

| Issue ID | Issue name | Category | Severity | Location | Net impact | Evidence | Description | Steps to reproduce | Suggested fix |
|---|---|---|---|---|---|---|---|---|---|
| #1 | Refresh rerenders entire cell | Performance & rendering at scale | Moderate | in-diff — `src/scenes/Editor/Notebook/cells/Cell.tsx:164` | Multi-statement dashboards spend up to 141 ms rerendering one cell. | Playwright React profiler: the same eight-statement refresh produced 13 renders/44.7 ms at head `a36aa4e`, versus 2 renders/6.5 ms at base `72ca5487`. | `CellInner` subscribes to every raw refresh-engine state object, although grid cells do not consume the subscribed value. Eight statements add eleven full-cell renders; 32 statements produced 37 renders and 141.4 ms of cell work. This repeats at the two-second auto-refresh minimum and compounds across dashboard cells. | 1. Create an eight-statement grid cell.<br>2. Run it, then refresh the unchanged results.<br>3. Count `CellInner` commits: head consistently produces 13 versus 2 at base. | Subscribe to a selector-filtered `fetching` boolean, suppress unchanged values, and avoid installing the subscription for grid cells—for example, isolate the draw Stop gate in a small child. |
| #2 | Divider collapses tall cell | Accessibility & UX | Moderate | in-diff — `src/scenes/Editor/Notebook/cells/useCellResizeOrchestration.ts:111` | Tall split-cell users lose 2,300 px in one divider interaction. | Scratch Vitest at head `a36aa4e`: 4,800 px becomes 2,500 px. Base End path is N/A because keyboard resizing is a new surface; the equivalent base mouse partition preserves the total. | With both panes at the permitted 2,400 px maximum, the middle divider advertises a 4,700 px maximum. Pressing End partitions the cell as 4,700/100, after which the independent pane clamp converts it to 2,400/100, unexpectedly collapsing the total height. | 1. Set a split SQL cell's editor and result heights to 2,400 px.<br>2. Focus the middle divider.<br>3. Press End.<br>4. Observe the allocation fall from 4,800 to 2,500 px. | Derive divider bounds so both panes stay within their individual maxima: clamp the requested top between `sum - MAX_PANE_HEIGHT_PX` and `min(MAX_PANE_HEIGHT_PX, sum - minBottom)`. Add the 2,400/2,400 boundary test. |
| #3 | Stop control untested | Test review & coverage | Moderate | in-diff — `src/scenes/Editor/Notebook/cells/CellStopButton.tsx:28` | Notebook runners lack regression coverage for the new cancel control. | Exact test search at head `a36aa4e` found no component or data-hook matches; lower-layer cancellation tests passed 141/141. Base is N/A because this is a new surface. | Existing tests verify cancellation helpers and the refresh engine directly, but none mounts the new control or verifies its visibility, grid/chart routing, click dispatch, removal, or focused-unmount handoff. Those integration points could regress while every current test remains green. | 1. Hold a notebook cell's first `/exec` request.<br>2. Assert that the Stop control appears and click it.<br>3. Assert cancellation and button removal.<br>4. Repeat for the initial chart-fetch branch. | Extend the existing delayed-`/exec` Cypress notebook flow to cover grid and chart cancellation, including focus handoff. |

## Validation

- `yarn typecheck`: passed
- `yarn build`: passed
- `yarn lint`: passed
- `yarn test:unit`: passed — 95 files, 2,160 tests
- `git diff --check`: passed
- Browser rung: the responsive chart zoom/reset path passed with Playwright against the no-auth `localhost:10000` server. The repository Cypress helper hard-codes an authenticated login, so that scenario was executed directly in the browser.
- Temporary review files and worktrees were removed. Pre-existing workspace changes were left untouched.

## Summary

- **Verdict: approve with comments** — address #1–#3, but none is blocking.
- Correctness gate: passed — no admitted Critical findings.
- Test gate: passed — 1 admitted Moderate coverage gap and 0 Critical coverage gaps.
- Findings: 3 in-diff, 0 out-of-diff. The level-3 cross-context inventory was rechecked and the remaining findings are contained within the changed surface.
- Severity: 0 Critical, 3 Moderate, 0 Minor.
- Adjacent findings: none.

PR #612, opened 2026-09-16
feat: query activity drawer

emrberk:
## Query Activity drawer

A new **Query Activity** button in the right sidebar, directly below Table Details, opens a drawer that lists the queries currently running on the server. It is built on the server's `query_activity()` function, which now reports live memory usage per query.

### What the drawer shows

**Header.** The title carries a badge with the number of running queries.

**Summary tiles.**
- **Running queries**: how many queries the server is executing right now.
- **Memory in use**: the total native memory currently held by those queries. Background work such as WAL apply and view refresh reports no memory and is not counted.

**Search.** The search box filters the list as you type. It matches the query text, the query id, and the username, case-insensitively. Pressing Escape clears the box. Pressing Escape on an empty box closes the drawer.

**Sort.** A dropdown offers eight orders: Duration, Memory used, Start time, and Query ID, each ascending or descending. The default is Duration descending, so the longest running query is on top. Queries with no memory data always sort last under Memory used.

**Auto refresh.** The refresh toggle polls the server once a second while it is on. Turning it off freezes the list. The choice is remembered across sessions. Toggling it either way fetches a fresh list immediately.

### What a query row shows

- **State**: an hourglass for *Running*, a cross for *Cancelled*, a check for *Finished*.
- **Memory**: the memory the query holds right now. When the query has a memory limit it reads `used / limit`. Background work shows `N/A`.
- **Duration**: how long the query has been running, updated every second. Hover it to see the exact start time with a copy button.
- **Query text**: the SQL, formatted for reading. Very long statements show their first and last lines with the middle elided. Hover the block for **Open in editor**, which opens the full formatted query in a new tab, and **Copy**, which copies the full formatted query.
- **Footer**: the user who ran the query, the worker pool and worker id, a **WAL** tag when the statement is being applied by the WAL job, the query id, and the **Cancel** button.

### Cancelling a query

Cancel asks for confirmation, then sends a cancel request to the server. The row switches to *Cancelled* and stays until the server drops the query, which happens at the query's next safe point. Cancel is not offered for WAL rows, because the server refuses to cancel them, nor for rows that are already cancelled or finished. On Enterprise, cancelling another user's query requires the SQL ENGINE ADMIN permission; the server's error is shown as a notification.

### Memory warnings

Only memory produces a warning or error tone, and only for queries that run under a memory limit. The tone reflects how much of that limit the query holds. Queries without a limit, and background work that reports no memory, are never coloured. Duration never colours a row, since what counts as long depends on the workload.

| Memory used | Tone | Hover text on the state icon |
|---|---|---|
| under 50% of the limit | none | — |
| 50% to 80% | warning | More than 50% of the available memory used |
| 80% and above | error | More than 80% of the available memory used |

Cancelled queries are never graded.

### How a query disappears

The server only lists queries while they run. It does not report an outcome. When a poll no longer returns a query, the row switches to *Finished* with the hover text "The query left the registry. It finished, failed, or timed out." The row stays for five seconds, then fades out.

While your mouse is over the row, or keyboard focus is inside it, the row stays as long as you like. When you leave it, a fresh five seconds starts. A cancelled query that leaves the server lingers the same way but keeps its *Cancelled* state.

### Other things to know

- On Enterprise, users without SQL ENGINE ADMIN see only their own queries. Administrators see everything.
- The drawer's own listing query is hidden from the list.
- Query ids are assigned by the server and start again from zero after a restart.
- If the server cannot be reached, the drawer shows an error. If it stops responding while the drawer is open, the last successful list stays visible with a banner until polling recovers.
- The list is virtualised and stays smooth with thousands of running queries.
- CPU time and disk I/O per query are not shown because the server does not expose them yet. The drawer is designed so they can be added as further columns.
- Finished queries are not kept beyond the grace period. Query history is a separate feature.

### Changes outside the drawer

Read-only SQL blocks elsewhere in the console, in the AI chat, the table details DDL, and the shared link confirmation, now render as static highlighted text instead of an embedded editor. They look the same, load faster, and their text can be selected with the mouse. Their copy button now appears on hover, like the open-in-editor action already did.


emrberk:
Reviewing PR #612 at level 3

## Issues

| Issue ID | Issue name | Category | Severity | Location | Net impact | Evidence | Description | Steps to reproduce | Suggested fix |
|---|---|---|---|---|---|---|---|---|---|
| #1 | Populated composer traps Escape | Cross-context caller impact | Moderate | in-diff — `src/components/Drawer/index.tsx:126-170,245-252` | AI-chat users lose Escape dismissal whenever a draft is present. | Cypress: drawer stayed open at `adf1d4fe`; identical test passed with the Drawer hunk reverted to `9b914793` behavior. | The shared Drawer now assumes every populated native text field clears itself on Escape. The AI composer handles Enter only, so Escape neither clears the draft nor dismisses the drawer; repeated presses remain inert. This regresses the base keyboard contract. | <ul><li>Configure and open AI chat.</li><li>Type an unsent draft in the composer.</li><li>Press Escape one or more times.</li><li>Observe that the draft and drawer remain open.</li></ul> | Keep Drawer’s default Escape-to-dismiss behavior. Make two-stage clear-then-close behavior an explicit consumer opt-in, or have only consumers that actually clear their field prevent the first Escape. |
| #2 | Toggle-off skips final refresh | Query execution & data integrity | Moderate | in-diff — `src/scenes/QueryActivity/index.tsx:115-120` | Operators can freeze Query Activity on a stale pre-toggle snapshot. | Independent Cypress: no listing request within 700 ms at `adf1d4fe`; N/A — new surface. | Turning auto-refresh off only persists the setting. It disables the adaptive loop without requesting the fresh snapshot promised for either toggle direction. If server activity changes after the last poll, the frozen list can omit that change indefinitely until the user re-enables refresh or reopens the drawer. | <ul><li>Open Query Activity with auto-refresh enabled.</li><li>After the current response, change the mocked/server query set.</li><li>Turn auto-refresh off before the next scheduled poll.</li><li>Observe that no immediate listing request occurs and the prior snapshot remains.</li></ul> | Trigger an explicit `fetchNow`/`refresh` for both toggle directions. Ensure the off transition’s final request is not canceled by disabling the polling loop. Add the missing on-to-off Cypress case. |
| #3 | Focused finished row persists | React correctness & hooks | Moderate | in-diff — `src/scenes/QueryActivity/useQueryActivity.ts:94-103,126-143`; `src/scenes/QueryActivity/QueryActivityRow.tsx:275-309` | Keyboard users leave focused finished queries visible indefinitely. | Independent Cypress at `adf1d4fe`: Finished row remained after 9 seconds; N/A — new surface. | When a focused running query disappears, the current snapshot first unmounts the row and a passive effect then reconstructs it as Finished. Unmount loses focus without releasing the parent’s held ID. Every later expiry tick therefore treats the row as held, so the advertised five-second grace period never starts unless the user happens to hover and leave the replacement row. | <ul><li>Open Query Activity with a running query.</li><li>Focus the row’s elapsed-time button.</li><li>Return a subsequent activity snapshot without that query.</li><li>Observe the replacement Finished row after more than five seconds; it remains indefinitely without further pointer interaction.</li></ul> | Preserve the row through the running-to-finished transition or release its held ID during unmount. Make the snapshot/finished transition atomic enough to avoid the intermediate removal, and add a focused-row expiry Cypress case. |
| #4 | Launcher focus not restored | Accessibility & UX | Moderate | in-diff — `src/scenes/QueryActivity/index.tsx:142-158` | Keyboard users lose their place whenever Query Activity closes. | Independent Cypress at `adf1d4fe`: visible launcher was not focused after close; N/A — new surface. | The visible `QueryActivityButton` opens the Redux-controlled drawer, but the Drawer registers an unrelated non-focusable `<span />` as its Radix trigger. On close, Radix attempts to restore focus to that span, so focus does not return to the control that opened the drawer. | <ul><li>Focus the Query Activity sidebar button and open the drawer.</li><li>Close it with the drawer’s close button.</li><li>Observe that the Query Activity launcher is not focused.</li></ul> | Wire the visible launcher to Radix as the actual trigger, or pass a launcher ref and restore it explicitly in `onCloseAutoFocus`. Add a keyboard focus-return assertion. |

None.

## Summary

- **Verdict:** **approve with comments** — address #1–#4, but none is merge-blocking under the review rubric.
- **Correctness gate:** passes; no admitted Critical issues.
- **Test gate:** passes; 2 admitted Moderate coverage gaps and 0 admitted Critical coverage gaps.
- **Quality checks:** `yarn typecheck`, `yarn build`, `yarn lint`, and `yarn test:unit` passed. Unit tests: 98 files, 2,158 tests.
- **Browser rung:** `e2e/tests/console/queryActivity.spec.js` passed all 15 tests. Targeted falsification artifacts reproduced each issue above; the live `query_activity()` SQL also succeeded against the bundled authenticated QuestDB instance.
- **Location split:** 4 in-diff, 0 out-of-diff. The shared-caller pass covered Drawer, LiteEditor/HighlightedSql, SQL formatting, adaptive polling, catalog sources, local-storage context, sidebar state, and the extracted Table Details utilities.
- **Severity distribution:** 0 Critical, 4 Moderate, 0 Minor.
- **Tradeoffs:** the drawer’s core listing, cancellation, sorting, error, and retry paths are well covered, and measured list recomputation remained below 8 ms at 10,000 rows. The remaining issues are bounded to toggle freshness and keyboard/focus contracts.

puzpuzpuz:
Reviewing PR #612 at level 3 (full mission-critical pass), against head `5bd1cfc3` and base `ec3b8bb6`.

**Verdict: request changes.** The new SQL formatter silently changes or breaks some valid QuestDB queries. That is the one Critical finding. The other eleven findings are ten Moderate and one Minor.

### PR title and description
- **Title:** it follows Conventional Commits. The repo convention repeats the verb, so it should be `feat: add query activity drawer`.
- **Description omits user-visible changes:**
  - The SQL formatter was swapped from `sql-formatter` to `@questdb/sql-parser`, so Format Document output changes for nearly every query.
  - There is a new "Capitalize keywords on format" editor setting.
  - Escape now behaves differently in every drawer.
- **Description claims the code doesn't meet:**
  - "Toggling it either way fetches a fresh list": turning auto-refresh off sends no request (browser-tested). The practical effect is at most one poll interval, so this is only a wording fix.
  - "If the server cannot be reached, the drawer shows an error": false on reopen (#6).
  - "Very long statements show … the middle elided": false for long single lines (#8).
  - Static SQL blocks "load faster": false for large SQL (#4).

## Issues

| ID | Issue | Category | Severity | Location | Net impact | Evidence | Description | Steps to reproduce | Suggested fix |
|---|---|---|---|---|---|---|---|---|---|
| #1 | Formatter splits QuestDB literals | Query execution & data integrity | Critical | in-diff: `src/utils/formatSql.ts:11`. Reaches unchanged consumers: editor Format (`questdb-sql/formatting.ts:9,27`), AI apply (`EditorProvider/index.tsx:829`, `aiAssistant.ts:562`), Query Activity copy/open | Users of hex/float/geohash literals get wrong results or broken DDL | Cypress with a real Alt+Shift+F plus the live server, at `5bd1cfc3` and at `ec3b8bb6` (base keeps every token intact) | The new formatter inserts a space inside single tokens: `0x1F`→`0 x1F`, `16777217.0f`→`16777217.0 f`, `geohash(5c)`→`geohash(5 c)`. Some results change silently: `SELECT id, 0x1F FROM t` runs green and returns INT 0 in a column named `x1F` instead of LONG256 `0x1f`. `…0f` returns a DOUBLE with a different value. Others fail: geohash DDL and casts (the documented precision syntax), `WHERE h = 0x…`, `VALUES(…, 1.5f)`. AI suggestions go through `normalizeSql` unconditionally, after the model has already validated the raw SQL. The capitalize setting doesn't matter. | <ol><li>Type `SELECT id, 0x1F FROM t`</li><li>Alt+Shift+F</li><li>Run; the grid shows 0</li></ol>Or format `CREATE TABLE g (h geohash(5c), ts timestamp) timestamp(ts)` → "invalid GEOHASH size". | Fix the sql-parser lexer so hex, `f`/`F`-suffixed numbers, `Nc`/`Nb` geohash precision and `Nns` stay single tokens, then bump the package. Add a formatSql corpus test that compares input and output with whitespace removed. For the AI path, fall back to the raw SQL if that check fails. Keep the new formatter's fixes for `!~`, `<<=` and `1.5E+3`, which base broke. |
| #2 | Saved notebook grid layouts orphaned | Persistence & migrations | Moderate | out-of-diff: `notebookColumnLayoutStore.ts:10` → `ResultGridPanel.tsx:49`, `GridShimmer.tsx:217`. Broken contract: the persisted key is derived from formatter output | Notebook users on 2.0.0–2.0.3: one-time layout reset | Round-trip vitest with the real base save and head load: base finds 18/18, head 4/18. Cypress with a seeded store: base applies the widths and pin, head doesn't | Layout keys hash `normalizeSql` output. The formatter swap changes that output for almost every query, so saved widths, column order and pins no longer load after upgrade. There is no migration or fallback lookup, and the old entries stay orphaned until LRU eviction. | <ol><li>On 2.0.3, resize or pin columns in a notebook cell</li><li>Upgrade</li><li>Reopen the notebook: default layout</li></ol> | Key on formatter-independent text (e.g. whitespace-collapsed SQL) so future bumps can't orphan layouts again. Either look up the legacy key once, or state the reset in the release notes. |
| #3 | SQL review hides bidi characters and line tails | Browser compat & security | Moderate | out-of-diff: share-link dialog `Monaco/index.tsx:2648`. Broken contract: LiteEditor render changed from a Monaco view to a static `<pre>` (`LiteEditor/index.tsx:396`, `HighlightedSql/index.tsx:17-19,54`) | Share-link recipients can't see disguised writes | Cypress, head vs base: base shows red `[U+202E]` markers and wraps long tokens; head shows neither (scrollWidth 1402 vs 516) | `colorize` runs with control-character rendering off, so bidi characters reorder the displayed text. A crafted link can make `DROP TABLE x` look like it sits after `--`, and confirming runs the DROP. Long runs with no spaces no longer wrap, so a statement glued after them is hidden behind a faint horizontal scroll. Mitigations: the generic write-statement warning, keyword colouring, and the editor behind the dialog all remain. | <ol><li>Open `/?query=<SQL with /*U+202E U+2066*/ DROP TABLE t; …>&executeQuery=true`</li><li>Compare the dialog text with the real SQL</li></ol> | Render U+202A–202E, U+2066–2069, U+200E/F, U+061C and C0 characters as visible markers in HighlightedSql. Add `overflow-wrap: anywhere` to its `<pre>`. |
| #4 | Large SQL freezes AI chat | Performance & rendering at scale | Moderate | out-of-diff: `AIChatWindow/index.tsx:1043`, `ChatMessages.tsx:705,791`, `AssistantMarkdown.tsx:151`. Broken contract: LiteEditor now renders every line | AI users with multi-thousand-line statements: 1–6 s freezes | Cypress timing, head vs base: 5k/10k/20k lines → 1.0/2.2/6.2 s longest frame gap at head; base flat at ~0.12 s | `maxHeight` is now just CSS. Every line is colorized synchronously (Monaco's colorizer is quadratic in lines) and inserted into the DOM, where base Monaco rendered about 11 lines. This repeats on every chat reopen and theme switch. | <ol><li>Paste a 10k-line INSERT</li><li>Click its AI glyph: the page stalls about 2 s</li></ol> | When `maxHeight` is set, colorize only the lines that fit, or render plain text above a line threshold. |
| #5 | Finished row sticks, focus lost | React correctness & hooks | Moderate | in-diff: `useQueryActivity.ts:126-137`, `QueryActivityRow.tsx:275-310`, `queryActivity.ts:232` | Keyboard and click users: finished rows never expire | Cypress at `5bd1cfc3` (MutationObserver plus commit probe): row node replaced, focus on `<body>`, row still present after 12 s. Hover-only control expires normally. | `finished` is updated in a passive effect, so the row unmounts for one commit and remounts as Finished. Focus inside it drops to `<body>`, and React 17 drops the blur, so the hold stays set. Held ids never expire and are never pruned. The same leak happens when a filter or an Escape-close unmounts a hovered row. | <ol><li>Click a running row's elapsed or copy button</li><li>Move the mouse away</li><li>The query ends</li><li>The Finished row stays until the user hovers it</li></ol> | Build the finished rows in the same render as the snapshot (derive them synchronously). Release the hold on row unmount, and prune `heldIds` to ids that are still listed. |
| #6 | Stale durations presented as current | Query execution & data integrity | Moderate | in-diff: `useQueryActivity.ts:51,69`, `queryActivity.ts:231` | Operators during outages see inflated or previous-session data | Cypress against the live server at `5bd1cfc3`: a 3 s query shown Finished "21s"/"34s". On a failing reopen, a previous-session row showed "Running 39s" (really 12.7 s) with Cancel | (a) Reopening during an outage never shows "Unable to load". The previous session's rows come back as Running, with durations inflated by the whole closed period, a stale badge and summary tiles, and only a "from the last successful response" banner. (b) A Finished row's duration runs to the next successful poll, not to when it was last seen. After auto-refresh off→on, or backoff, it shows start + gap, and sorts accordingly. | <ol><li>Open, then close the drawer</li><li>Make the server fail</li><li>Reopen: old rows show as Running</li></ol>Or: auto-refresh off, wait 20 s, turn it on → a 2 s query shows ~22 s. | Reset `lastReadyData` per drawer session so a failed reopen shows the error. Freeze the Finished duration at the last-seen snapshot. |
| #7 | Table Details shows "Unavailable" after one blip | Cross-context caller impact | Moderate | in-diff policy change `TableDetailsDrawer/index.tsx:260,306-327` (MANUAL policy plus deadline re-evaluation) affecting the unchanged `DetailsTab.tsx` | Table Details users hit ~2 s false "Unavailable" | Cypress, head vs base: head 2.05 s "Unavailable" with Copy DDL disabled; base ~1 s "Loading…", or data immediately | One failed SHOW COLUMNS or DDL request now makes the source unavailable at once (grace 0). A policy switch re-judges earlier failures, and POLLING then needs two successes to recover. | <ol><li>The first SHOW COLUMNS fails on open (Monitoring tab)</li><li>Open Details: "Unavailable", Copy DDL disabled for ~2 s</li></ol> | Don't re-judge counted failures when the policy changes, or keep POLLING for columns/DDL. Fetch immediately when Details becomes active. |
| #8 | Long single-line query row unbounded | Performance & rendering at scale | Moderate | in-diff: `queryPreview.ts` (line-count elision only), `QueryActivityRow.tsx:365` (no `maxHeight`) | Rows can reach ~8,000 px | Cypress plus the live server: a real 24 KB query rendered as one 8,491 px row; a no-space literal gave a 144,000 px horizontal scroll | Elision counts lines only, and the server doesn't truncate query text. Inlined literals or long expression chains fill the drawer for ~19 screens. | <ol><li>Run `SELECT '<20 KB text>' …`</li><li>Open Query Activity</li></ol> | Also cap the preview by characters per line, or give the row's LiteEditor a `maxHeight`. |
| #9 | Escape swallowed by non-clearing fields | Accessibility & UX | Moderate | out-of-diff: `Drawer/index.tsx:126,165` affects `TableSelector`, the CSV import Settings and Schema dialogs, and the main editor with any side drawer open | Keyboard users: Escape does nothing in several contexts | Cypress, head vs base: base closes on the first Esc; head stays open after three | The guard treats every non-empty input or textarea as one that "clears itself", which is false outside the two search boxes. It matches read-only inputs, untouched prefilled fields, and Monaco's hidden textarea. Mixed impact: at base, Esc in the AI composer and import forms also discarded the draft, so head protects those. | <ol><li>Open Table Details</li><li>Shift+Tab to the table-name field</li><li>Press Esc repeatedly: nothing happens</li></ol> | Make clear-then-close opt-in (e.g. a `data-clears-on-escape` attribute) and fix the comment. |
| #10 | Weak and missing Query Activity tests | Test review & coverage | Moderate | in-diff: `queryActivity.test.ts:183,258`, `queryActivity.spec.js` | Regressions in cancelled grading and error banners go unnoticed | Mutation run: with the `state === "cancelled"` branch removed, 27/27 unit tests and the e2e logic still pass. Grep: zero tests for `query-activity-stale` and the cancel-error toast. | Every cancelled fixture has no memory limit, so it grades "none" regardless of state. Cancelled rows with a limit are a real server state. The stale banner and the cancel-error toast (e.g. the Enterprise permission error) are untested. | Remove the cancelled branch in `classifyQuery`; all tests stay green. | Add a cancelled fixture with a 1 GiB limit and 900 MiB used. Add e2e tests for the stale banner (success, then 500) and for a cancel returning 400. |
| #11 | Cancel dialog drops keyboard focus | Accessibility & UX | Moderate | in-diff: `CancelQueryDialog.tsx:18` | Keyboard and screen-reader users lose their place | Cypress (real keys) at `5bd1cfc3`: `activeElement` is BODY after Keep or Escape | The AlertDialog has no trigger, so Radix restores focus to nothing. The same pattern exists in older dialogs, but this instance is new. | <ol><li>Tab to a row's Cancel button, press Enter</li><li>Press Esc: focus is on `<body>`</li></ol> | Save the opener when the dialog opens and focus it in `onCloseAutoFocus`. |
| #12 | Dead fade constant, wrong comment | Code structure, readability & types | Minor | in-diff: `queryActivity.ts:54-58` | Developer-facing; rows vanish ~0.7 s early | N/A: static, grep shows no uses | `FINISHED_FADE_DELAY_MS` is unused, and the comment describes a transition delay that `QueryActivityRow.tsx:79` doesn't apply. | — | Apply it as `transition-delay` or delete it, and fix the comment. |

## Adjacent findings
None met the evidence bar.

## Summary
- **Verdict:** **request changes**, because of #1.
- **Correctness gate:** fails on #1, an open Critical. Every admitted finding had its strongest disproof attempted by an independent falsifier, with an executed artifact at `5bd1cfc3`. Where applicable (#1, #2, #3, #4, #7, #9) the same trigger was also run at base `ec3b8bb6`.
- **Test gate:** passes. There are no Critical coverage gaps; #10 is the one Moderate test finding.
- **Quality checks:** `yarn typecheck`, `lint`, `test:unit` (98 files, 2,158 tests) and `build` all pass. `queryActivity.spec.js` passes 15/15 against a live QuestDB 10.0.2-SNAPSHOT.
- **Location split:**
  - Out-of-diff breakage (4): #2, #3, #4, #9.
  - In-diff with reach into unchanged consumers (2): #1, #7.
  - New Query Activity surface (6): #5, #6, #8, #10, #11, #12.
- **Severity:** 1 Critical, 10 Moderate, 1 Minor.
- **Tradeoffs:**
  - The new formatter is much faster than the old one: a 558 KB query formats in 70 ms vs 53 s.
  - It fixes several shapes the old one broke (`!~`, `<<=`, `1.5E+3`).
  - Static rendering is lighter than Monaco for typical short SQL.
  - Backoff doesn't affect existing consumers: the Table Details and notebook fetches never reject.
- **Earlier review comment on this PR:** its Escape (#9) and focused-finished-row (#5) findings still reproduce at head, as does the toggle-off gap (description note above).

emrberk:
Reviewing PR #612 at level 3 (head `42c3baad`, base `ec3b8bb6`).

## PR title and description
- **Title:** repeat the verb: `feat: add query activity drawer`.
- **Claims the code does not meet:**
  - "The drawer closes on the next [Escape]": false for fields that do not clear themselves (#1).
  - "A spinner appears … while a refresh is in flight": it never shows for Retry or the refresh after a cancel (#8).
  - "The exact start time with a copy button": the time is cut to milliseconds (#7).

## Issues

| ID | Issue | Category | Severity | Location | Net impact | Evidence | Description | Steps to reproduce | Suggested fix |
|---|---|---|---|---|---|---|---|---|---|
| #1 | Escape swallowed in drawers | Accessibility & UX | Moderate | out-of-diff: `Drawer/index.tsx:126-171` affects AI chat, Import settings, Table Details, and the main editor | Keyboard users: Escape never closes the drawer from a filled field | Playwright, 3 presses per context: head stays open in all 4 contexts; base closes on the 1st press | Any focused input or textarea with a value blocks Escape. The value never clears, so the next press does not close the drawer either. This includes Monaco's hidden textarea, a read-only title input, and prefilled import fields. The AI chat draft kept at head is the only real gain. | 1) Open Table Details. 2) Click into a non-empty SQL editor. 3) Press Esc 3 times: the drawer stays open. | Make clear-then-close opt-in, for example a `data-clears-on-escape` attribute on the two search inputs. |
| #2 | Share-link dialog hides SQL | Browser compatibility & security | Moderate | out-of-diff: `Monaco/index.tsx:2651` (`scrollable`), `HighlightedSql/utils.ts:1-10` | Share-link recipients can confirm SQL they cannot see | Playwright: at head, "DROP" is at x=3516 and the visible area is x=554–1046, with no scrollbar. Base wraps it into view. Zero-width characters are marked at base and not at head | `white-space: pre` keeps text after a long run of spaces off-screen. U+200B/C/D, U+2060–2064 and U+00AD are no longer marked. The write warning and the plural title still show. | 1) Open `?query=SELECT 1;` + 400 spaces + `DROP TABLE t;&executeQuery=true`. 2) The dialog shows only `SELECT 1;`. | Drop `scrollable` and use `pre-wrap` with `overflow-wrap: anywhere`. Add zero-width and format characters to `CONTROL_CHARACTER_RANGES`. |
| #3 | Table Details "unavailable" after one blip | Cross-context caller impact | Moderate | out-of-diff: `TableDetailsDrawer/index.tsx:310-325` (MANUAL policy) affects the unchanged `DetailsTab` | Table Details users: about 2 s of false "unavailable", Copy DDL disabled | Playwright plus a state-machine unit test: head shows "unavailable" for 2.0 s; base shows "loading" for 1.0–1.2 s, then the data | While Monitoring is active, 1 failed SHOW COLUMNS or DDL fetch makes the source unavailable at once. On Details, the source needs 2 successful polls to recover. | 1) The first SHOW COLUMNS fetch fails. 2) Click the Details tab: "unavailable" for about 2 s. | Keep POLLING for columns and DDL, or fetch at once when Details becomes active. |
| #4 | Saved notebook grid layouts lost | Persistence & migrations | Moderate | out-of-diff: `notebookColumnLayoutStore.ts:10` → `ResultGridPanel.tsx:49`, `GridShimmer.tsx:217` (the key comes from formatter output) | Notebook users upgrading from 2.0.x: one-time layout reset | Vitest round-trip with the real base and head stores: 10 of 12 keys differ; a base save gives `null` on a head load | The formatter swap changes the key for almost every query. Saved widths, order and pins no longer load. No fallback exists. | 1) On 2.0.3, resize a notebook grid column. 2) Upgrade: the default layout returns. | Use formatter-independent text as the key (whitespace collapsed), or look up the legacy key once. At the least, note it in the release notes. |
| #5 | Finished row never expires | React correctness & hooks | Moderate | in-diff: `QueryActivityRow.tsx:307-312, 343-348` | Mouse users who copy the start time | Playwright: the row stays Finished at +5/15/30/45 s; focus is on BODY | Copy in the portaled tooltip sets `focused=true`. The tooltip then unmounts, and React 17 drops the blur, so the hold is never released. | 1) Hover "Started … ago", then click copy. 2) Move the mouse away. 3) The query ends: the row stays. | Release the hold on the row's `focusout` with a native listener, or track focus with `:focus-within` when the pointer leaves. |
| #6 | Long single-line query row | Performance & rendering at scale | Moderate | in-diff: `queryPreview.ts:20`, `QueryActivityRow.tsx:365` | Rows about 6 screens tall for queries with big literals | Playwright: a real 17 KB query gives a 6,371 px row; a literal with no spaces gives a 144,520 px scroll | The preview counts lines only, and the server does not truncate `query`. | 1) Run a query with a 17 KB string literal. 2) Open Query Activity. | Also cap characters per line in `buildQueryPreview`, or give the row's LiteEditor a `maxHeight`. |
| #7 | Start-time tooltip loses data and access | Accessibility & UX | Moderate | in-diff: `QueryActivityRow.tsx:268, 343-361` | All rows: time cut to ms; copy only with a mouse | Playwright: server `…58.409078Z`, clipboard `…58.409Z`. Enter/Space/tap closes the tooltip; Tab never reaches copy | `new Date().toISOString()` drops microseconds, although `formatUtcTimestamp` keeps them. The plain Tooltip closes on activation. `MonitoringTab` already uses `useToggletip` for the same pattern. | 1) Hover the start time and copy it. 2) Or press Enter on it: the tooltip closes. | Use `formatUtcTimestamp` and `useToggletip`. |
| #8 | Manual refresh shows no spinner | Async, timers & cancellation | Moderate | in-diff: `useQueryActivity.ts:139-169` | Users who click Retry or cancel on slow servers get no feedback | Playwright, 2 runs: a slow poll shows `aria-busy=true`; a slow Retry and a slow post-cancel refresh stay false for 2.8 s | `isFetching` is one boolean. The restarted poll returns "skipped", or `fetchNow` aborts it, and its `finally` clears the flag. | 1) Auto refresh on, and the listing is slow. 2) Click Retry: no spinner. | Use a counter, or track only `fetchNow`. Do not bump `pollGeneration` in `refresh`. |
| #9 | Cancel dialog drops focus | Accessibility & UX | Moderate | in-diff: `CancelQueryDialog.tsx:17-28` | Keyboard users lose their place | Playwright with real keys: after Keep or Esc, focus is on BODY; the next Tab leaves the drawer | The dialog has no Trigger, so Radix returns focus to nothing. | 1) Tab to Cancel, press Enter. 2) Press Esc. | Store the opener and focus it in `onCloseAutoFocus`. |
| #10 | Cancel, then reopen: endless loading | Async, timers & cancellation | Moderate | in-diff: `index.tsx:149-150`, `useCatalogSource.ts:132-136` | Auto refresh off and slow server: rare, but no recovery path | Playwright with a delayed CANCEL: loading from +1 s to +20 s. Control run without the delay loads normally | The `refresh` from the earlier render aborts the reopened drawer's request. Its own response is then skipped. With auto refresh off, the loading state has no Retry button. | 1) Auto refresh off, then confirm Cancel. 2) Close and reopen the drawer before CANCEL returns. | Read `refresh` through a ref, or skip it when the session generation changed. |
| #11 | Capitalize setting stale across tabs | State & context architecture | Moderate | in-diff: `formatSql.ts:7-12` | Users with several tabs: the toggle does not match the format result | Playwright, 2 tabs: B formats in uppercase while its switch shows off; saving B turns the setting off everywhere | `formatSql` reads localStorage directly, while the provider reads the value once. At base, settings took effect through provider state, so a stale tab still matched its own UI. | 1) Tab A: turn capitalize on. 2) Tab B: format, then save another setting. | Pass `capitalize` from `useLocalStorage` to the formatter callers, or listen for `storage` events. |
| #12 | Weak and missing tests | Test review & coverage | Moderate | in-diff: `queryActivity.test.ts`, `HighlightedSql/utils.test.ts`, `queryActivity.spec.js`, `editor.spec.js` | Regressions in grading, security markers and error states pass CI | Mutation runs on scratch copies: removing the cancelled branch passes 26/26; removing U+202A–D, U+2067–8, U+200F or C1 passes 4/4. Grep finds no test for `query-activity-stale` and no check on the dialog SQL text | Cancelled fixtures have no memory limit. Bidi fixtures test only range endpoints. The stale banner and the share-link SQL text have no test. | Remove the cancelled branch in `classifyQuery`; the tests stay green. | Add a cancelled row that has a limit, a test for each range, e2e for success then 500, and a check on the dialog SQL text. |
| #13 | Small code cleanups | Code structure, readability & types | Minor | in-diff | Developer-facing only | N/A, static: grep and source lines | `FINISHED_FADE_DELAY_MS` is unused, and its comment is wrong. A second `escapeHtml` repeats `utils/escapeHtml.ts`. `describeMemoryStatus` does not use `row`. Inline `style` at `QueryActivityRow.tsx:346`. `1.2rem` instead of `theme.fontSize.xs`. The PR adds 7 comments. The search input has no `aria-label`. | — | Delete or apply the constant. Reuse the existing `escapeHtml`. Remove the unused parameter. Use theme tokens. Remove the non-critical comments. |

## Adjacent findings
None met the evidence bar.

## Summary
- **Verdict:** approve with comments. Please fix #1, #2, #3, #4 and #5 before merge. The other items can follow.
- **Correctness gate:** passes; there are no admitted Critical findings.
- **Test gate:** passes. There are no Critical coverage gaps; #12 is the one Moderate gap.
- **Quality checks:**
  - `typecheck`, `lint` and `build` pass.
  - `test:unit` passes (99 files, 2,161 tests).
- **Location split:** out-of-diff 4 (#1–#4), in-diff 9.
- **Severity:** 0 Critical, 12 Moderate, 1 Minor.
- **Tradeoffs:** the description already records two costs of the static SQL rendering as known limits, so I did not count them as findings:
  - Large SQL in the AI chat freezes the page for 0.9 s at 10k lines and 2.1 s at 20k lines. At base the freeze is about 0.1 s.
  - Long tokens in DDL and AI chat blocks are clipped, with no visible scrollbar.
- **Gains:**
  - Format Document runs much faster: 57 ms against 73 s for a 568 KB INSERT.
  - At the realistic maximum row count, which connection limits cap near 320, sorting costs only a few milliseconds per second.

